PR #3, opened 2020-09-04
G1000 import

kaosfere:
Fairly straight forward.   The extra thing I did  here was modify the build script so that the package name  was optional, and if not provided it will build each package in the project.   I figured this would make it easier for us to do single-zip releases, but if you don't like it I can revert that.

I also changed the project file names a bit so it's easier to keep them separate from the package files.  Too much xml!


mattnischan:
LGTM!

PR #7, opened 2020-09-04
WIP: tag-based release automation

kaosfere:
This modifies the workflow for the G1000 to cuz a release when a tag with a g1000-v* prefix is pushed to the repo.  There are two different versions, one for pre-releases and one for full releases.  Currently I have the releases being set as drafts to we can review them until there's confidence in the automation.

This does not yet have anything for creating a text description on releases, I'm thinking about the best way to handle that.

I'm creating this as a WIP PR to get comment on what I've done, and also see if you want me to include similar steps for the CJ4.

You can see the outputs of this process here:   https://github.com/Working-Title-MSFS-Mods/fspackages/releases

### DIFF
--- a/.github/workflows/workingtitle-g1000-build.yml
+++ b/.github/workflows/workingtitle-g1000-build.yml
@@ -1,4 +1,4 @@
-name: CI - G1000
+name: CI - G1000 Build
 on:
   push:
     branches:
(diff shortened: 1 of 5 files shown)


kaosfere:
(This has a lot of small commits I had to make because it's impossible to test CI automation without actually pushing anything to the repo.   This should probably be a squashed merge when it goes in.)

PR #13, opened 2020-09-05
removed directory filter from copies in build script

kaosfere:
The filter for directories only in the file copy here made it skip any files in the top level of the package directory.  This will avoid that.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -46,12 +46,12 @@ foreach($packageEntry in $projectFile.Project.Packages.Package) {
     Write-Host "Copying source files..."
     foreach($assetGroup in $packageDef.AssetGroups.AssetGroup) {
         $dest = Join-Path $packagePath $assetGroup.OutputDir
-        foreach($assetSubFolder in Get-ChildItem -Path $assetGroup.AssetDir -Attributes Directory) {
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -46,12 +46,12 @@ foreach($packageEntry in $projectFile.Project.Packages.Package) {
     Write-Host "Copying source files..."
     foreach($assetGroup in $packageDef.AssetGroups.AssetGroup) {
         $dest = Join-Path $packagePath $assetGroup.OutputDir
-        foreach($assetSubFolder in Get-ChildItem -Path $assetGroup.AssetDir -Attributes Directory) {
+        foreach($assetSubFolder in Get-ChildItem -Path $assetGroup.AssetDir) {
```
mattnischan, line 49: The reason `Directories` was here was to avoid problems below with the recursive copy on non-folders.

kaosfere: Not quite sure I follow.   That's the only recursive copy in this from what I can see.  Below, you recurse to look for files to put in the manifest, but you don't do any copying.

kaosfere: for the permanent record, discussed in channel and we're good with this.

PR #19, opened 2020-09-08
merging G1000 work back to main

kaosfere:
This brings my G1000 code from pre-release to main for full release.  I'll squash this when I merge this... there were so many changes to the build workflow itself that I had unfortunately to push a lot of in-progress work the origin so the log is pretty hairy.

This doesn't actually touch anything outside of the content I've been working on had tested by a few folks in pre-release.


mattnischan (approved):
Looks a little beefy from a pulling in whole files standpoint, but with the way Asobo has things I don't see really any alternative, to keep change exposure to a minimum.

kaosfere:
> Looks a little beefy from a pulling in whole files standpoint, but with the way Asobo has things I don't see really any alternative, to keep change exposure to a minimum.

Nope, it's kinda grody.  It would be really nice for an instrument to be able to reflect upon its own configuration.

PR #30, opened 2020-09-10
Retrofit g3000 - removed as requested the bonanza and the skylane from the metadata file

argonox:
tbm930_g3000_inhpa
In regard to the name "gx":
The idea behind that is, that on one side, it is not realy one g1000 anymore and on the other side, the garmins are very dependend due its "share" folder files. When it comes to the build process it could raise problems, when we handle it seperatly.

How do you think about it?

best regards,
Argonox


kaosfere:
Looks awesome, thanks a lot!

PR #32, opened 2020-09-12
Fix false positive "DON'T SINK" alerts

tavip:
The MaxSinceStart XML element is used to compute the "Altitude Loss After Takeoff" by subtracting the current altitue from the maximum altitude since start. This approach triggers a false TERRAIN / DON'T SINK alert when flying circuits / patterns or when performing touch and go landings.

To fix this issue reset the maximum value of the MaxSinceStart XML element when we are on ground.


kaosfere:
This looks great, thanks.   This was bugging me just the other day when I was doing touch-and-goes in the DA40 and it felt wrong but I hadn't looked into why yet. :)   

I'm going to merge this into a new branch I'm creating for integrating changes to the next release, and this will go out with that.   

Thanks again!

PR #34, opened 2020-09-14
Reimplement fastToFixed function to include trailing zeros.

### LINKED ISSUE #21: altimeter setting
Like the GPS used to, the altimeter setting will show 30.1 or 29.9 instead of 30.10 and 29.90.

tavip:
Fixes #21 

Before this patch:

```
fastToFixed(12.345, 4) -> "12.345"
fastToFixed(12, 1) -> "12"
```

After this patch:

```
fastToFixed(12.345, 4) -> "12.3450"
fastToFixed(12, 1) -> "12.0"
```

I also tested the performance impact of the new implementation and it is
about twice slower than the original one, however it is now correct and
much better then toFixed() which is about 20 times slower.

The tests were performed with the following script in Chrome
(JavaScript	V8 8.5.210.20):

```
a=performance.now();
for(let i=1; i<10000000; i++) {
    fastToFixed(12.345, 1);
}
b=performance.now();
b-a
```

Before patch: 132.78999998874497
After patch: 230.62000000476837
toFixed: 1958.7400000020862

### DIFF
--- a/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/CommonPFD_MFD.js
+++ b/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/CommonPFD_MFD.js
@@ -491,11 +491,7 @@ class PFD_NavStatus extends NavSystemElement {
                     this.legSymbol = 2;
                 }
             }
-            var rawCurrentLegDistance = fastToFixed(SimVar.GetSimVarValue("GPS WP DISTANCE", "nautical miles"), 1)
-            if (rawCurrentLegDistance.indexOf(".") == -1) {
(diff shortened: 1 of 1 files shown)


kaosfere:
This looks great.   Have you done any head to head comparisons to see if that performance difference has any notable effect on framerate or render time?

tavip:
@kaosfere I haven't seen any difference in FPS or CoherentGTDraw when using the stock simulator (no g1000 package) or between g1000 package with or without my patch. 

Since the initial code is using toFixed() in a few places (e.g. the baro display) which is way slower I don't think it should cause any real performance degradation for G1000 aircraft.

I did notice that there are a few calls when drawing the XML gauges for non g1000 aircraft, but in that case we should not overwrite the fastToFixed function since we do that from common_pfdmfd.js. Just to make sure, I have also tested with and without the patch with c172 classic and did not notice any FPS and GTDraw differences.

kaosfere:
Sounds solid to me.  I'll go ahead and merge this to the dev branch.   Thanks!

PR #37, opened 2020-09-15
Build.ps1: Implement Filewatcher for automatic builds

dga711:
When the build.ps1 is now run with the argument `-WatchFiles $true` it will watch for file changes in the `src` folder.
On a file change, it will automatically run the build routine.

+ Use robocopy to only copy modified files to build output
+ New parameter `$CleanBuild` to clean the output path before build

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -3,7 +3,8 @@ param (
     [string]$Package,
     [string]$MinimumGameVersion = "1.7.14",
     [string]$OutputPath = ".\build\",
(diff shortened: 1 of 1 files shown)


kaosfere:
LGTM.   It would be cool to enhance this so it only changes files that have been updated rather than rebuilding the whole thing, but that's another iteration.

mattnischan (approved):

kaosfere (approved):

PR #39, opened 2020-09-15
Fixed issue where TAKEOFF REF QNH would accept bad values.

mattnischan:


kaosfere (approved):

cwburnett:
@mattnischan I could do another PR for landing, but it might be faster if you don't mind making the same change for the landing QNH as well?

mattnischan:
> @mattnischan I could do another PR for landing, but it might be faster if you don't mind making the same change for the landing QNH as well?

Yeah, I can add that to this one.

mattnischan:
@cwburnett Added to the approach page now also.

cwburnett (approved):

PR #40, opened 2020-09-15
Light mod for the CJ4. Will apply only apply to CJ4, not globally.

Uwajimaya:


kaosfere (approved):

cwburnett:
@Uwajimaya I hate to do this, but is there a chance you could make whatever compatibility adjustment you made for your packages to this one as well for the lights? Want to be sure when we roll this out in beta in the next couple days that the lights part is working ;-) If not, I can probably track it down by looking at your original code, but if you could update, it would be a great help! Thanks!

Uwajimaya:
It's actually super easy - and you can do it yourself. The release package needs to include a manifest.json - make sure that manifest.json is current - just copy and paste from... \Microsoft Flight Simulator\Packages\Official\Steam\asobo-aircraft-cj4. That's it. 
There are no changes to my specific lighting files.

cwburnett:
@Uwajimaya perfect, thanks, since we have a manifest for the workingtitle-aircraft-cj4 package then we should be good! Thanks for the quick reply!

PR #47, opened 2020-09-17
Added remaining landing performance calculations

K20017:


cwburnett (approved):
Looks good to me, let's have @mattnischan or @kaosfere take a quick look for syntax and then we can merge.

mattnischan (approved):
I think these definitely need to be split out into a calcs class, but we can get to that later.

cwburnett:
Merging with Main. In agreement we need to try and pull all the performance calcs into their own class.

PR #49, opened 2020-09-17
CJ4: Activate yaw damper & adjust landing gear warning conditions

musurca:
-activates YD (yaw damper) in autopilot;
-landing gear aural warning follows training manual behavior more closely.

Currently I have implemented the following valid scenarios that should produce an audible warning when the gear is not down:

- flaps > 15-degrees, both throttles < CRU
- flaps <= 15-degrees, RADALT < 500 ft
- flaps > 15-degrees, one engine failed but not set to STOP, opposite throttle < CRU
- flaps <= 15-degrees, one engine failed but not set to STOP, opposite throttle < CRU, RADALT < 500ft

I have NOT implemented the following valid scenarios that should also produce an audible warning:
- Landing gear test systems active, both throttles < CRU
- flaps <= 15-degrees, RADALT invalid, and AOA > 0.4
- flaps <= 15-degrees, one engine failed but not set to STOP, RADALT invalid, and AOA > 0.4

These could be addressed in the future, but currently I'm not sure which Simvars indicate a test of the landing gear, or how to determine whether the RADALT is "invalid" (which I assume means that the device is damaged or inoperative?).  

(Acknowledgements: YD fix, systems.cfg for gear warning, and training manual selection per forum thread: https://forums.flightsimulator.com/t/citation-cj4-yaw-damper-gear-warning-fixes/272098)


cwburnett:
@musurca first, thank you so much for this. Do you know if these changes are negatively impacted by the Asobo update at all? Will this need to be edited, or is this set of changes good to go with the latest version?  Thanks!

musurca:
ah, thanks for catching that. I diff'ed the Asobo CJ4 files after the latest patch and brought in the two lines that had changed. (With the exception of @Uwajimaya's lighting changes, which I left in.) Just tested it and everything seems to be working.

cwburnett (approved):

mattnischan (approved):

PR #54, opened 2020-09-18
CJ4: Reenable stepping through legs on MFD via FMC

panourgue:
Very small tweak that re-enables "stepping" through "legs" one-by-one in the MFD map via FMC. 

I have no idea how it's like IRL, but in a game I use this functionality all the time :)

In pre-1.8.3.0 you could step through legs one-by-one by enabling "rose" map mode on PFD. In this patch they changed it from "rose" to "plan" mode (based on the names of the variables, they actually use Indices, "3" to be exact). But there is no such thing in PFD, so I switched it to use MFD instead.

step through legs

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_LegsPage.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_LegsPage.js
@@ -157,7 +157,7 @@ class CJ4_FMC_LegsPage {
             }
         }
         fmc.currentFlightPlanWaypointIndex = offset + step + 1;
-        let isMapModePlan = SimVar.GetSimVarValue("L:CJ4_MAP_MODE", "number") === 3;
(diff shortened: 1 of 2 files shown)


cwburnett:
@panourgue thanks for this, I'll check this against what I'm working on with LEGS. IRL the step should work only on PLAN and only on the MFD, and I think I have that working in a draft on my machine, but I'll double check later today. Thanks for submitting this PR!

panourgue:
> @panourgue IRL the step should work only on PLAN and only on the MFD

That's exactly what my commit does

mattnischan:
@panourgue Did you mean to close this PR?

panourgue:
@mattnischan 
Yeah, turns out it's broken. Why do you ask?

mattnischan:
We were just looking at PRs and were looking to pull this in for the release, and we were going to give it a test.

Were you planning on looking at it further or did you want to just leave it closed? It's totally OK to leave the PR open while you investigate a fix, if you like.

panourgue:
I don't have time this week to work on msfs. And it looks like @cwburnett works on this anyways, so...

mattnischan:
OK, no worries.

Thanks for your contribution nonetheless! Happy to consider further PRs in the future.

PR #58, opened 2020-09-19
G1000 engine display

knighty:
Added Engine Display page
Changed PFD and MFD styling
Added flight plan marker


src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.html
```diff
@@ -0,0 +1,968 @@
...
+                <div id="SubText">
+                    Press the FMS CRSR knob to return to base page
+                </div>
+            </div>
+        </div>
+    </div>
+</script>
+
+<link rel="stylesheet" href="AS1000_MFD.css" />
+<link rel="stylesheet" href="AS1000_MFD_Custom.css?53" />
```
kaosfere, line 929: Let's knock that query string off of here.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.js
```diff
@@ -13,10 +13,11 @@ class AS1000_MFD extends BaseAS1000 {
         });
         this.trackup = false;
         this.pagesContainer = this.getChildById("RightInfos");
-        this.addIndependentElementContainer(new Engine("Engine", "LeftInfos"));
+        this.engineDisplay = new Engine("Engine", "LeftInfos");
+        this.addIndependentElementContainer(this.engineDisplay);
         this.pageGroups = [
             new NavSystemPageGroup("MAP", this, [
-                new AS1000_MFD_MainMap()
+                new AS1000_MFD_MainMap(this.engineDisplay)
```
kaosfere, line 20: I get that what you're doing here is trying to separate then engine logic from the map element so increase the encapsulation.   One one hand, that's a good idea.  On the other hand, it kinda breaks the (admittedly sloppy) paradigm that they wrote this stuff with.

On the whole, since we're supposed to be fixing their broken shit, I'm OK with the change. :D

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD_Custom.css
```diff
@@ -0,0 +1,145 @@
+@font-face {
```
kaosfere, line 1: What's your thinking behind having these as separate files from the main ones?  Is it just to avoid messing with the stock elements as much as possible?   I  haven't been too worried about that, since like I said, we're fixing stock anyway...   but this could make our lives easier especially when it comes to them updating things.   And if by doing this you don't have to commit an entire CSS file in one go that's also a win.

I *think* I'm OK with this, but I reserve the right to change my mind on that.  :D   Confirm your logic here and I think we can talk about it with the group on Slack.  We're trying to maintain a somewhat consistent paradigm in how we work across projects.

kaosfere (changes requested):
This looks pretty good on the whole, made a couple comments above that need addressing.

The main thing in my mind at this point is whether we can adjust this so it doesn't break if you load a plane that doesn't have an engine display configuration.      If possible I'd love for people to be able to use the "core" of the mod and gracefully fall back to sensible defaults if they choose not to use our panel.

We can totally ship panels for the stock planes...   but I wouldn't want to do anything which will make this _completely unusable_ for commercial planes that use the G1000.     At it stands right now there is no way I could use this with the Carenado 182 for example.  and that would make me sad.

Any thoughts on how we could accomplish that?

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.html
knighty: Sorted

kaosfere (changes requested):
Cool.   I think just rename the new CSS files from _Custom.css to _WT.css (or _Engines.css, whatever you feel better about) and we can mergey-merge

kaosfere (approved):
DO EET! :)

PR #62, opened 2020-09-20
add manual handling (ew) for the avionics knobs of popular commercial…

kaosfere:
I feel a little gross about this, but I'm tired of telling folks that this won't work 100% with the 182.


knighty:
This seems reasonable to me unless there's a way we can think of to just improve this system in general. Do we disable the menu if we lack the knob index? If not I figure that it something we should do.

knighty (approved):
Looks fine to me

kaosfere:
I'm not going to close this one unmerged...   once the experimental config loader is pulled in, this becomes unnecessary.

PR #69, opened 2020-09-20
Altimeter: show correct barometric pressure when switching unit

tavip:
When switching the barometric pressure unit (from inches to HPA and
viceversa) the shown barometric pressure value is reset to 29.92IN
(or HPA equivalent).

Save the barometric pressure updates in the local thisPressure field
and use it to update the HTML text either when the pressure changes or
when the unit changes.


kaosfere (approved):
I like the way you've done this, it's concise and elegant without being overly confusing.   Thanks for this one, too!

PR #73, opened 2020-09-21
implement dynamic configuration loading

kaosfere:
This is the final result of the discussions we had today and my day of work on it.   This implements a class with several file load functions, the two main ones being

a) loading a .cfg file into a dict
b) loading an .xml file into an XML DOM document

It also has convenience functions to just load a raw file, as well as shortcuts for loading an interior or exterior model.xml without needing to manually open and examine the model.cfg to figure out where they are.

As we noticed today, the xml files that make up the interior model definitions, at least, are not actually valid XML, so there's an additional step in place when loading those to slap a fake root round them so the document is valid XML.

This also includes two avionics config files, one for the G36 and one for the C172 to demonstrate the configurability.   With this version of the mod installed, you should be able to load the G36 and have PFD at max dim , MFD at max bright, and north up on the map.   Then hop into the 172 and you'll have PFD at max bright, MFD at max dim, and track up on the map.

Both of there planes are also being read for their avionics brightness pot so the onboard knob works for them too without needing any external configuration.

There is a fair bit here, and it's something that I think could be useful in other projects too, so I'm making this a WIP as a discussion prompt. Let me know what y'all think!


src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.js
```diff
@@ -1108,7 +1125,7 @@ class AS1000_MapMenu {
                 }
             case "TRCK UP":
                 {
-                    if (this.gps.trackup)
+                    if (this.trackup)
```
kaosfere, line 1128: Oops, I think this was a mistake.  I'll check tomorrow.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
```diff
@@ -0,0 +1,103 @@
...
+                section = match[1];
+                ptr = out[section] = out[section] || {};
+                return;
+            }
+            
+            let key = match[2];
+            let value = match[3];
+            if (["true", "yes", "on", "1"].includes(value)) {
+                value = true
+            } else if (["false", "no", "off", "1"].includes(value)) {
```
kaosfere, line 93: oops, that should be a 0 there not a 1

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
```diff
@@ -0,0 +1,103 @@
...
+            cb(out);
+        })
+    }
+
+    // The model xml files aren't properly formed because they have multiple root
+    // nodes. (Seriously?)   We'll wrap them in a fake root node so that we can
+    // parse them properly.  Hopefully that's actually a consistent format.
+    loadFrenchXml(filename, cb) {
+        Utils.loadFile(`${this._vfspath}/${filename}`, (text) => {
+            text = text.replace(/\<\?.*\?\>/, '');
```
kaosfere, line 31: this ugly-ass regex is to remove the XML DTD.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
```diff
@@ -0,0 +1,103 @@
...
+    parse(cfgString) {
+        var out = {};
+
+        var ptr = out;
+        var section = null;
+
+        var re = /^\[([^\]]*)\]$|^([^=\s]+)\s*= *([^#\s]+)/i;
+        var lines = cfgString.split(/[\r\n]+/g);
+
+        lines.forEach((line, crap, morecrap) => {
```
kaosfere, line 72: you can't seem to use _ as a throwaway variable name in this environment

tavip:
@kaosfere I was working on another approach to make settings persistent, using the existing dataStorage APIs. Please see #75

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
```diff
@@ -0,0 +1,103 @@
+class ConfigLoader {
+    // This class kind of became callback hell without me meaning for it to.
+    // TODO:  make it less of a callback hell
+    constructor(basepath) {
+        let vfspath = "/VFS/" + basepath.replace(/\\/g, "/");
+        let vfspathParts = vfspath.split("/");
+        this._vfspath = vfspathParts.slice(0, -2).join("/")
+    }
+
+    loadCfg(filename, cb) {
```
mattnischan, line 10: If you wanted this to be slightly less callback madness, you could use promises here:

```javascript
return new Promise((resolve, reject) => {
  Utils.loadFile(`${this._vfspath}/${filename}`, (text) => {
    let parser = new CfgParser()
    let out = parser.parse(text);
    resolve(out);
  });
});
```

Don't forget semicolons, have a few missing after your `loadFile` calls in this class. ;-)

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
mattnischan: It's JS, so you can just leave the params out entirely: `() =>`

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
mattnischan: Are `True` and `False` also valid? Just thinking about case sensitivity.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.js
```diff
@@ -33,16 +33,27 @@ class AS1000_PFD extends BaseAS1000 {
         this.addEventLinkedPopupWindow(new NavSystemEventLinkedPopUpWindow("Procedures", "ProceduresWindow", new MFD_Procedures(), "PROC_Push"));
         this.addEventLinkedPopupWindow(new NavSystemEventLinkedPopUpWindow("CONFIG", "PfdConfWindow", new AS1000_PFD_ConfigMenu(), "CONF_MENU_Push"));
         this.maxUpdateBudget = 12;
-        let avionicsKnobIndex = 30;
-        let avionicsKnobValue = SimVar.GetSimVarValue("A:LIGHT POTENTIOMETER:" + this.avionicsKnobIndex, "number");
+        this.avionicsKnobIndex = 30;
+        this.avionicsKnobValue = 0; // SimVar.GetSimVarValue("A:LIGHT POTENTIOMETER:" + this.avionicsKnobIndex, "number");
+        this._interiorCfgLoaded = false;
```
mattnischan, line 38: Could we make a `AS1000Config` class that attaches here instead that holds this state and talks to the config backend?

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
kaosfere: Thanks.   I tried mucking with something along those lines but couldn't get it sorted.  I could give it another go though.  You have more time in the javascript mindset than I do;  for this environment do we gain enough by using promises instead of callbacks?    Callbacks fit better into my mindspace but I don't want to write pointlessly ugly code and async JS is pretty new to me.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.js
kaosfere: 1)   I need 'line' in the loop there so I can't use empty params.   I tried using _ for the throwaways which I was pretty sure _shoud_ work, but it didn't.

2) I was thinking of throwing `value = value.toLowerCase()` in this block for just that reason.  I'll go ahead and add that.

3) For sure, this section needs a heavy rewrite with @tavip's recent work merged in.   I plan on doing just that, creating a config class which pulls from simvars or the data storage and provides a cleaner interface to the mian code.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
kaosfere: (Just to follow up for the Permanent Record(tm),  we discussed this on the side and I'll spend a little time reworking this with promises.)

kaosfere:
OK, at this point I think this about how I'd want it to be.   I feel like the initialization logic in the PFD and MFD is now streamlined enough that there isn't really a lot of commonality between them to where it makes sense to add a config-holder class just for the G1000.

I think what it *might* be good to do is to move the PersistVar logic out of CommonPFD_MFD.js, in pursuit of the discussions we've had about minimizing our footprint, and put it into the new workingtitle-core (or workingtitle-lib or whatever) package we were talking about, then create a higher level logic wrapper around it in the same place that all displays can use.  Then the 1000 or the 3000 or even the Proline can use it.

I feel like that is outside the scope of this particular PR though.  It would definitely be good to do _at least_ before the G1000 v0.4, and perhaps also before the next non-bugfix CJ release.

But I welcome thoughts on that as always! 👍

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.js
```diff
@@ -33,19 +33,65 @@ class AS1000_PFD extends BaseAS1000 {
...
+        // overridden by the new position of the knob.  The only way I have found to set this is,
+        // unfortunately, to stick a little pause in here to let the systems load a bit before
+        // completing the initialization of our current settings.  It's kinda gross, but since this
+        // is an async function it's not *too* bad, and the worst it means it that the hardware
+        // knob, if present, won't function for a few seconds.
+        //
+        // The 10000ms value here was experimentally derived by testing to see what caused loads in
+        // troublesome planes to succeed on my system.  It may need adjustment if other folks find
+        // themselves having that issue. 
+        await new Promise(resolve => setTimeout(resolve, 10000));
```
dga711, line 58: Check if it would help subscribing to onFlightStart that is available in some base classes of the sim.
Or just check the use of the gamestates.

see BaseInstrument.js
```
    onFlightStart() {
        console.log("Flight Starting...");
        SimVar.SetSimVarValue("L:HUD_AP_SELECTED_SPEED", "Number", 0);
        SimVar.SetSimVarValue("L:HUD_AP_SELECTED_ALTITUDE", "Number", 0);
        this.dispatchEvent(new Event('FlightStart'));
    }
    onGameStateChanged(_oldState, _newState) {
        if (_newState != GameState.mainmenu) {
            this.createMainLoop();
            if (_oldState == GameState.loading && (_newState == GameState.ingame || _newState == GameState.briefing)) {
                this.reboot();
            }
            else if (_oldState == GameState.briefing && _newState == GameState.ingame) {
                this.onFlightStart();
            }
        }
        else {
            this.killMainLoop();
        }
        this._gameState = _newState;
    }
```

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.js
```diff
@@ -33,19 +33,65 @@ class AS1000_PFD extends BaseAS1000 {
...
+                this.setBrightness("MFD", avionicsKnobValueNow);
+                this.avionicsKnobValue = avionicsKnobValueNow;
+            }
+        } 
+    }
+
     onEvent(_event) {
         if (_event == "MENU_Push") {
+            // Uncomment this for easy debugging of config issues.
+            // this.pfdConfig();
```
dga711, line 94: You can use
`g_modDebugMgr.AddDebugButton("InsertTitleHere", function() {dosomething});`

This will always show a button in the devkit panel that can do stuff

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.html
```diff
@@ -929,6 +929,8 @@
 <link rel="stylesheet" href="AS1000_MFD_WT.css" />
 
 <script type="text/html" import-script="/JS/SimPlane.js"></script>
+<script type="text/html" import-script="/JS/dataStorage.js"></script>
```
tavip, line 932: While looking at some other stuff, I have noticed that we may be able use JS imports like this:

```
Include.addScript("/JS/Services/Aircraft.js");
```

That would allow us to pull dataStorage.js in CommonPFD_MFD.js and avoid changing every *.html files we want to use it and also remove the try/catch blocks in PersistVar.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.js
kaosfere: That's a really great idea, and I've played with it a bit to try to make it work...  but I'm afraid it goes back to the same unreliable behavior I was trying to fix with that sleep.  :(   The problem is that we need to wait until the model interior xml has been loaded and parsed before we can assign a value to the avionics knob and _that_ needs to be done before we can flag the PFD ready for control.   But if we do this in onFlightStart we can't await the load there because it's not an async function.  So we go back to square one.   I'll keep thinking about other ways around this.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.js
kaosfere: Ah, sweet.  I remember that from the first time I read the devkit docs but I'd never used it then forgot.  :D  Thanks for the reminder.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.html
kaosfere: Interesting idea and I think I like it.  Any thoughts on that, @dga711 , @mattnischan ?

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.js
kaosfere: This works now by waiting for a promise in onFlightStart

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
```diff
@@ -0,0 +1,119 @@
+class ConfigLoader {
```
mattnischan, line 1: I think the only two small things I have in this file are:
* I seem to remember you changed CfgParse to IniParser, so I just want to make sure I'm looking at the right stuff
* The methods and classes all need javadoc.

mattnischan (commented):
Looks good! Just one small comment above.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/ConfigHandler.js
kaosfere: I changed loadCfg to loadIni to more accurately represent the particular file format it's meant to parse.  The class loads both ini and XML files (both "normal" XML and the warped kind Asobo are using with multiple roots) so "Config" felt, to me at least, like it remained more appropriate there.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.html
kaosfere: opened an issue for us to come back to that, thanks, @tavip!

mattnischan (approved):
Kickin'. 🔥

src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/MFD/AS1000_MFD.html
dga711: > Interesting idea and I think I like it. Any thoughts on that, @dga711 , @mattnischan ?

Yeah. Actually. I was thinking about also putting this as a comment to include it that way. But then I was like "eh, you have changed the html already anyway" ^^
But yeah, @tavip is right

mattnischan: The only downside I can see is that it might not allow us to cache bust as easily.

dga711: @mattnischan luckily addscript runs through the same hoops as the includes in the script-tags.

PR #75, opened 2020-09-21
Persistent PFD settings

tavip:
This adds support for persistent (per plane) PFD settings using the existing dataStorage APIs.

The following settings are currently implemented: wind data mode, reference speeds values and enable/disable state, bearing source and DME visible / not visible, barometric pressure unit.

### DIFF
--- a/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/Shared/BaseAS1000.js
+++ b/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/Shared/BaseAS1000.js
@@ -927,12 +927,24 @@ class AS1000_PFD_AirportInfos extends NavSystemElement {
 }
 class AS1000_PFD_AirspeedReference {
     constructor(_valueElem, _statusElem, _refSpeed, _displayName) {
-        this.isDisplayed = false;
+        this._isDisplayed = PersistVar.get(`VSRef.${_displayName}.IsDisplayed`, false);
         this.valueElement = _valueElem;
         this.statusElement = _statusElem;
         this.refSpeed = _refSpeed;
-        this.displayedSpeed = _refSpeed;
+        this._displayedSpeed = PersistVar.get(`VSRef.${_displayName}.DisplayedSpeed`, _refSpeed);
         this.displayName = _displayName;
+    }
+    get displayedSpeed() {
+        return this._displayedSpeed;
+    }
+    set displayedSpeed(val) {
+        this._displayedSpeed = PersistVar.set(`VSRef.${this.displayName}.DisplayedSpeed`, val);
+    }
+    get isDisplayed() {
+        return this._isDisplayed;
+    }
+    set isDisplayed(val) {
+        this._isDisplayed = PersistVar.set(`VSRef.${this.displayName}.IsDisplayed`, val);
     }
 }
 class AS1000_PFD_TMRREF extends NavSystemElement {
--- a/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.html
+++ b/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/PFD/AS1000_PFD.html
@@ -664,6 +664,7 @@
 <link rel="stylesheet" href="AS1000_PFD_WT.css" />
 
 <script type="text/html" import-script="/JS/SimPlane.js"></script>
+<script type="text/html" import-script="/JS/dataStorage.js"></script>
 
 <script type="text/html" import-template="/Pages/VCockpit/Instruments/Shared/Map/MapInstrument.html"></script>
 
--- a/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/CommonPFD_MFD.js
+++ b/src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/Shared/CommonPFD_MFD.js
@@ -13,6 +13,40 @@ class PFD_VSpeed extends NavSystemElement {
     onEvent(_event) {
     }
 }
+class PersistVar {
+    static get(key, defaultValue, prefix) {
+        var storeKey = `${SimVar.GetSimVarValue("ATC MODEL", "string")}.${key}`;
+        try {
+            var stringValue = GetStoredData(storeKey);
+            if (stringValue == null || stringValue == "")
+                return defaultValue;
+        } catch (e) {
+            return defaultValue;
+        }
+        switch (typeof defaultValue) {
+            case "string":
+                return stringValue;
+            case "number":
+                return Number(stringValue);
+            case "boolean":
+                return Boolean(stringValue);
+        }
+        return defaultValue;
+    }
+    static set(key, value) {
+        var storeKey = `${SimVar.GetSimVarValue("ATC MODEL", "string")}.${key}`;
+        switch (typeof value) {
+            case "string":
+            case "number":
+            case "boolean":
+                try {
+                    SetStoredData(storeKey, value.toString());
+                } catch (e) {
+                }
+        }
+        return value;
+    }
+};
 class PFD_Airspeed extends NavSystemElement {
     constructor() {
         super();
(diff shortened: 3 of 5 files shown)


mattnischan:
Hi @tavip,

Thanks for this and your other contributions! We're definitely going to take a look at all of them.

For the DataStorage APIs, we had discovered those as well, but hadn't gotten to testing them yet. Have you found that these values persist between flights? Or is the storage fresh for each flight?

kaosfere:
Yeah, when I saw this this morning my jaw dropped.   Is those persist... that's exciting.

tavip:
The settings persists between flight and simulator restarts. I have done PC reboots as well to be sure and they are still there. I've been digging though the local files to see where they are saved but haven't figured it out where yet.

The bearing/dme stuff is having some issues and is still work in progress, but the rest seems to work fine for me.

kaosfere:
So, we just tested and yep, it absolutely persists, that's killer.   So, for this PR:  I absolutely *love* the concept, and I'm so happy you knew about this and shared it with us.  I do think that having the same boilerplate code at every data location is probably a bit overly heavy.

What I was thinking might be good would be to create a settings manager class somewhere in, perhaves, Navsystems.js so it's widely available.   This would deal with the logic of setting and safely retrieving individual values, so then it's just one function call in every instrument to get a value (if set), and another to set on change.

You have thoughts on that?   Would you be interested in adding that as part of this?

tavip:
A settings manager class definitely seems like a good approach, I'll rework the PR to add that. I would like to look at a few other settings that we want to persist to make sure the API is general enough, what are some other settings I should be aware of?

For example, initially I though that we could use getter/setters (like for AS1000_PFD_AirspeedReference), but it does not appear to be a viable approach for some settings.

knighty:
I'm actually working on settings stuff currently, involving adding a whole settings page to the MFD but part of that work was the creation of a class to handle managing the settings which gets dependency injected to places that need it (as best as we can given the structure of the code anyway). 

```javascript
class AS1000_Default_Settings {
}

AS1000_Default_Settings.base = {
    weight: "kg"
};

class AS1000_Settings {
    constructor(aircraft, defaults) {
        this.aircraft = aircraft;
        this.defaults = defaults;
        this.settings = {};
        this.lastUpdated = 0;

        this.load();
    }
    getValue(key) {
        if (key in this.settings) {
            return this.settings[key];
        } else if (key in this.defaults) {
            return this.defaults[key];
        } else {
            return null;
        }
    }
    setValue(key, value) {
        this.settings[key] = value;
    }
    getStorageKey() { 
        return "config|" + this.aircraft;
    }
    getTimestampKey() { 
        return "config_timestamp|" + this.aircraft;
    }
    save() {
        SetStoredData(this.getStorageKey(), JSON.stringify(this.settings));
        this.lastUpdated = (new Date()).getTime();
        SetStoredData(this.getTimestampKey(), this.lastUpdated);
    }
    load() {
        let storedData = GetStoredData(this.getStorageKey());
        if (storedData) {
            this.settings = JSON.parse(storedData);
            this.lastUpdated = GetStoredData(this.getTimestampKey());
        }
    }
    update() {
        let lastUpdated = GetStoredData(this.getTimestampKey());
        if (lastUpdated && lastUpdated > this.lastUpdated) {
            this.load();
        }
    }
}```

This is basically what I've been working on. Kind of unfortunate we both ended up doing this is seems haha. The `update` is there because we need to keep settings objects in sync across panels

**Comment by kaosfere** [comment]
@knighty, I see what you're going for.  You're looking at storing all the settings in a single json doc, whereas @tavip is thinking more along the lines of individual keys.   I think either way could be good, you're just trading off where you keep the complexity -- in the storage layer, or in the stored data itself.

I'm torn on which way would be better.   The one thing I am fairly confident think it would be good to build for would be each element managing its own settings.   Meaning, for example, when the PFD loads, the code for the altimeter reaches out to the settings manager to load the setting for baro-mode via a getKey call, and when updated saves it via the setKey call.   This is as opposed to having the PFD load the full setting doc and configure everything in one central location.   That could be a little tricky to maintain.

Aside from that, I think either storage layer approach (discrete keys vs a single doc) is valid and has its particular merits.

Feelings on that (or anything else)?

**Comment by kaosfere** [comment]
(Either way, if @tavip is interested I'm more than happy to let him finish the implementation.  :D  It's definitely a very sold start. )

**Comment by knighty** [comment]
Each element managing its own settings I would say is much more complex than just having a single repository for keys, especially as it'd be hard to predict where various settings might affect parts of the UI. As an example, the settings page I'm working on has the ability to change units which would be used all over the place.

Regarding individual keys vs a single key stored with json I agree there's merit to both approaches, I think think dumping everything in json is an easier way to handle having more complex structures if we need them down the line. If we're stuck with just strings then we're going to have to json encode anything that isn't just a string anyway, so putting the whole thing in there is of little consequence imo. Furthermore, if settings get removed there's high potential for dangling data we can't get rid of easily.

**Comment by kaosfere** [comment]
> As an example, the settings page I'm working on has the ability to change units which would be used all over the place.

So, for this, the middle-path that I see here would be, when someone changes the units, your settings page setting a global `${ATC MODEL}.global_units` through the settings manager.   Then, any instrument that wanted to know what settings it should use would get that key through the settings manager and configure itself appropriately.

That _does_ mean that if you had multiple instruments that cared about that setting changed we'd need to have some way of having "subscribers" that would know when a setting changed.   I think that is an interesting problem, and would be a good thing to have.  But, more importantly, I also think that's a bit beyond the basic settings manager itself and I don't want to overload this PR with a global pub/sub system, or talk about implementing that one through a data storage layer that's not meant to be a real-time multicast system.

(If we want to do that, we can use the LVar data bus concept. :D )

I'm feeling like we should decide on the back end format and then proceed with the PR as it exists with the devices that don't require, effectively, a message bus taking care of them selves.   If we want to expand beyond that for a full-fledged settings page changing the settings in multiple instruments that's another PR.  :)

**Comment by kaosfere** [comment]
Ok, the team talked about this a little bit more on the side.  I'm good with this as you have presented it.  I'd say let's go ahead with the rework you were thinking of to implement the manager class and we'll happily merge that.   As long as it provides an encapsulated interface to a storage layer with a simple getter and setter for individual keys we can enhance it in the future as much as we want.

What you have here on its own is a huge win and will definitely go out in the next release to the gratitude of many.  😁 Thanks a lot, and sorry for the sprawling debate.

**Comment by tavip** [comment]
No worries, thanks for the debate, I know there are many things to consider to make things maintainable and extendible in the future :) I'll update the PR in a few hours with the class that abstracts the storage layer and with fixes for the DME and bearing settings.

**Comment by tavip** [comment]
@kaosfere  @knighty  It look longer than I expected to fix the bearing / dme stuff and a bit more code to deal with updating the attributes as it seems one can not change them from the constructor - I'm a newbie when it comes modern web programming :) 

Other than that, the code now looks much better with the dedicated storage class API, let me know what you think.

File: src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/Shared/BaseAS1000.js [hunk 1]
```diff
@@ -927,13 +927,25 @@ class AS1000_PFD_AirportInfos extends NavSystemElement {
 }
 class AS1000_PFD_AirspeedReference {
     constructor(_valueElem, _statusElem, _refSpeed, _displayName) {
-        this.isDisplayed = false;
+        this._isDisplayed = PersistVar.get(`.VSRef.${_displayName}.IsDisplayed`, false);
```
kaosfere, line 930: This line has the one small quibble I have remaining.  I like the dot-notation for namespacing, but do you think it would be better to just append that automatically to the ATC MODEL in the persistence handler so that you don't need to with that first dot every time you're manipulating a variable?  If there's a reason you went with this way instead I'm not opposed to it, mostly curious from an ease-of-use perspective.

kaosfere (commented):
This looks good.   There's room to grow on it but this gives us a great start and I like it a lot more than needing to manage config files.  😀   Let's just settle on the question of where to do the first dot in the namespace and then I'm happy merging this.

tavip:
Good point! (no pun intended) Definitely easier to use the way you suggested, I'll fix it and repush.

kaosfere (approved):
Fabulous, thank you so much!

Do you have a user ID at Avsim or the MSFS foums you'd like credited?  I want to make sure everyone who helped gets a proper shout-out.  :)

kaosfere:
(Also, if there's anything from our issues you'd like to grab, or an enhancement of your own you want to pitch, let us know.  There's plenty of stuff for us to do.  😀 )

tavip:
Thanks for the reviews! I'll take a look in the issues and pick something :)

PR #80, opened 2020-09-21
Added recall swap and correct input parsing to TUNE page.

mattnischan:


src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
```diff
@@ -143,6 +143,26 @@ class CJ4_FMC extends FMCMainDisplay {
 
         //Timer for periodic page refresh
         this._pageRefreshTimer = null;
+
+        //Hijaack and amend the standard FMC logic to work with the PL21 TUNE page
+        let initRadioNav = super.initRadioNav.bind(this);
```
dga711, line 148: Why does this not fit into CJ4_FMC_NavRadioPage.js ? .bind(fmc) there should give you the same scope.
Could give it a static init? 
I am just not sure about this big block being in the init for CJ4_FMC

dga711: Hm. Seeing the full file now. Maybe just extract it to a method/function.

mattnischan: Yeah, I'm not a big fan of having to do this either, but we're sorta forced to here.

Moving to it's own method is a good idea.

dga711 (approved):

PR #91, opened 2020-09-22
[CJ4]: Remove reverse throttle for keyboard and gamepad users

panourgue:
Without this, keyboard/gamepad users are still able to reverse throttle using the "decrease throttle" game mapping (defaults: "B" for Xbox controller and F2 for keyboard) if they push the button so throttle decreases past zero, so to speak. Even though the plane moves slightly forward. Go figure ¯\\_(ツ)_/¯
Pretty sure it doesn't affect engine performance. At least numbers are the same.


mattnischan (approved):
Hi @panourgue,

Thanks for the contribution! Nice catch, we hadn't seen that.

PR #92, opened 2020-09-22
New setTemplate override

### LINKED ISSUE #46: CJ4: FPLN title and MOD handling
1. The title needs to be left aligned
2. It will change when FPLN is modified to show "MOD FPLN". I think the "MOD" part becomes white

image

dga711:
**TODO:**
- [x] use setMsg in pages
- [ ] strip out legacy code after full conversion of templates
- [x] fix height distribution

**FILES - TODO:**
- [x] DepArrPage (K)
- [x] DirectToPage
- [x] DsplMenuPage
- [x] FuelMgmtPage
- [x] IdentPage (K)
- [x] InitRefIndexPage
- [x] LegsPage (K)
- [x] MfdAdvPage
- [x] NavRadioPage
- [x] NavRadio Page 2 and 3
- [x] PerfInitPage
- [x] PosInitPage (K)
- [x] RoutePage
- [x] SelectWptPage
- [x] TakeoffRefPage

## Description
This PR offers overrides to the setTemplate functions of FMCMainDisplay.
It introduces a new class WT_FMC_Renderer, which contains
- ~New `setTemplate`, `setTitle`, `setLabel`, `setLine`~
- Char grid! Characters are now evenly distributed with a fixed width independent of font size.
- `showExec`, `hideExec`
- `setMsg`
- `renderSwitch`

~_CJ4_FMC_ has a new property `messageBox `which can be used to set the msg line from the pages like `fmc.messageBox = "MSG TEST"`~

There is also a new css file WT_FMC.css to have our own styles separate. Here you can also define new styles for the templates if needed.

## New setTemplateRaw
~You will still call `setTemplate` via `fmc.setTemplate` as today.~

Be aware that every whitespace is now rendered and is also used for Identation.
For example left-bound headers often are idented, so you also have to put that whitespace into the field's text.
Right-bound elements need the whitespace at the end of the text to indent.

### Conversion of legacy templates
As examples for conversion see TUN, IDX, FPLN.
Maybe read up on the new formatting syntax below.

#### 1. Change call
During conversion you will change the `fmc.setTemplate` call to `fmc._templateRenderer.setTemplateRaw`. 
When we are done and I strip out the legacy system, these will be renamed.

Per default it will render the rows with font sizes like small/normal/small/normal.... If you dont want that, add a `false` flag as second parameter to the call. (see _DepArr Page 1_). This could change if we think its stupid.

#### 2. Fix header line
The new format for the header line is `["LEFT","RIGHT","CENTER"]`
Before there was a 4th and 5th field for pagination. These are gone.
Most headers are blue so you need to add the `[blue]` format.

Example header:
`["", "", "DEP/ARR INDEX[blue]"]` (centered, no pagination)
`["", "1/2[blue] ", "INDEX[blue]"]` (centered, with pagination)
`[" ACT FPLN[blue]", "1/" + pageCount.toFixed(0) + "[blue] ", ""]` (left-bound with ident)

#### 3. Fix content lines
Just follow the new formatting syntax. 
Most cases will be to change a `[color]blue` to `[blue]`.

Remember whitespace for indentation and format `["LEFT","RIGHT","CENTER"]`

### Header

~The override provides new syntax for styling titles, labels and lines. Everything is backwards compatible, so if you leave the line as they are today, it will still work.~ Old syntax is not compatible when you switch to `setTemplateRaw`

~Titles (first line in setTemplate) can now contain five fields to support alignment and new styling syntax 
`["left", "right", "center", page, pagecount]`~

A `[]` can now contain a css class which is then applied to previous text. One line can contain multiple `[]` One bracket can contain multiple css classes.
Most used classes are the colors like: blue, green, yellow, pink . s-text for smaller text.

`"YES[green]/[white]NO[white s-text]"`
leads to 
image
the resulting HTML is like 
`<span class="green">YES</span><span class="white">/</span><span class="white s-text">NO</span>`
_Note: There is a `[white]` after the `/` so it is not in the span with `s-text`_

**Backwards Compat**
Title lines with three fields will function as of today.
For lines and labels it will resort to the old logic when it finds `[color]` bracket.

closes #46


dga711:
Did do basic stuff to some pages as the new templating allows it now
image
image
image
image
image

cwburnett:
note to @mattnischan don't approve this yet, we're going to commit all the template changes into this PR/Branch before merging

cwburnett:
@dga711 is there a way to build in some error correcting on the new template? Right now it is crashing when more than 24 chars are fed to a line, which is happening with a couple airport names that are longer than 24 chars. The real unit just cuts them down to 24 and doesn't show the rest. Can we adjust the template to always do that, or do we need to build in that trimming for each entry that could go long?  Thanks!

cwburnett:
@cwburnett fix lat/lon and about page

dga711:
@cwburnett done

cwburnett:
@dga711 @cwburnett @mattnischan @K20017 I think we are very close to being able to merge this into Main, but it would be great if we could each take a trip to the CJ4 cockpit sometime today and click through every page on the FMC to be sure we haven't missed any. K20017 and I did this last night and fixed a few things, but now that DGA has updated the template, it would be worth doing one more time with a larger audience. We're specifically looking for (a) any template problems that cause incorrect display and (b) any pages that cause an error (black screen, UNTITLED, erroneous data display or console errors).

Thanks!

src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_DepArrPage.js
```diff
@@ -185,12 +185,13 @@ class CJ4_FMC_DepArrPage {
             fmc.refreshPageCallback = () => fmc.onDepArr();
         };
         //end of CWB EXEC handling
+        modStr = fmc.fpHasChanged ? "MOD[white]" : "ACT[blue]";
 
-        fmc.setTemplate([
-            [originIdent + " DEPARTURES", currentPage.toFixed(0), pageCount.toFixed(0)],
-            ["DEPARTURES" + "[color]blue", "RUNWAYS" + "[color]blue"],
+        fmc._templateRenderer.setTemplateRaw([
+            [" " + modStr + originIdent + " DEPART", currentPage.toFixed(0) + "/" + pageCount.toFixed(0) + " [blue]"],
```
knighty, line 191: I would use template strings here as I think they're a lot easier to mentally parse
```javascript
` ${modStr}${originIdent} DEPART`
```
I've left the comment here but it's a running theme.

dga711: I wanted to use them, but I think I've tried it once and it didnt work.

But it could be that I used escapes
image

cwburnett (dismissed):

K20017 (commented):
It looks fine and running well on my tests.

cwburnett (approved):

PR #98, opened 2020-09-25
Fix nose up/dn button behavior during FLC (closes #89)

### LINKED ISSUE #89: G1000 autopilot FLC mode Nose Up/Nose Down reversed
I didn't know if it was going to be fixed in the patch (I reported it more than once to MSFS Zendesk - with acknowledgement) or if its in the scope of what y'all are doing with the G1000 but...

Exactly as the title says!

To reproduce: Climb out in any G1000 aircraft. Pre-select an altitude higher than you are. Hit FLC (it will capture FLC with your current airspeed, this is accurate). Press Nose Up. In the real aircraft, nose up means your airspeed goes down, because the nose of the airplane goes up. In the sim, Nose Up increases the airspeed lowering the nose. This is reverse of reality. Using VS, nose up will increase positive vertical speed raising the nose, FLC does the same (except in the sim where its backwards!).

Also regarding FLC, it cannot be set below the lowest operating airspeed of the GFC700 configuration for that aircraft. For example, in the G36 if you aren't moving and hit FLC it will set 80kts and cannot be set lower. If I remember its 70kts for the DA40NG, either way its in the POH under the minimum airspeed for autopilot operation listed with the specific airspeed. with FLC, you can't set the speed below minimum, even if you are currently stationary.

Last, would a curator of all these suggestions and organizing them into one space be a good idea? 
I liked how we had just one thread where people would add to the list. Perhaps we can have an organized master list of everyone's suggestions which y'all can triage into what's easy/important, what's specific to only 1 or 2 G1000 flavours (meaning the G1000s from different airframe manufacturers... like everything beechcraft being in Celcius). I can see a challenge in trying to sort through the mud of all this, people adding suggestions that only exist in the nxi version (which isn't modeled at all in the sim, or from what I understand you guys are doing), or specific to one model, etc.

kaosfere:
this seems to be limited to just the G1000.


mattnischan (approved):

knighty:
Looks good 👌

knighty (approved):

PR #100, opened 2020-09-25
RFC: persistent ATC options

tavip:
This PR saves the ATC options in between simulator sessions. I am not sure if this is the proper way for saving these settings since flight data seems to be driven from in game and not UI / HTML context and will probably be fixed later by Asobo. However this fix/workaround it might be useful in the meantime since it is a rather annoying issue. Also this does not belong in the g1000 package. But I am posting as it is for discussion.

### DIFF
--- a/PackageDefinitions/workingtitle-g1000-metapackage.xml
+++ b/PackageDefinitions/workingtitle-g1000-metapackage.xml
@@ -57,6 +57,14 @@
 			</Flags>
 			<AssetDir>docs\workingtitle-g1000</AssetDir>
 			<OutputDir>.\</OutputDir>
-		</AssetGroup>  		   
+		</AssetGroup>
+        <AssetGroup Name="workingtitle-base-ui-pages">
+			<Type>Copy</Type>
+			<Flags>
+				<FSXCompatibility>false</FSXCompatibility>
+			</Flags>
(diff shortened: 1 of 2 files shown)


knighty:
Yeah definitely got a G1000 thing but kick ass find, I'm sure a lot of people would be happy to see this fixed. Kind of a tricky problem here, you don't wan't to spread out tiny mod fixes across hundreds of packages but also it doesn't really belong in any larger one. Are you considering releasing it as a standalone mod?

tavip:
Do you think it makes sense to create a new package here? I know that it is probably out of scope for this project, but it would probably have a better chance of reaching more people and be better maintained if it would be release from a larger project :)

knighty:
I think the best course of action while the community is in this state of infancy is a bespoke package. I think there's too much danger for unmaintainable conflicts otherwise. I think long term things will probably setting into a handful of overhaul packages

kaosfere:
Hey @tavip , I didn't want to leave you hanging on this one.   I like this and definitely would be interested in seeing this evolve, but I don't really feel like it belongs with the G1000.   We chatted a little this evening, and I'm going to set up a new `workingtitle-experimental-ui` project for this to get pulled into where we can work on this and other neat features that don't really fit under the Garmin rubric.

tavip:
This issue seems to be fixed in 1.9.3.0, so I am going to close this PR.

PR #101, opened 2020-09-25
Fixed missing nav radio method separation from last PR and added periodic radio page refresh.

mattnischan:


cwburnett (approved):

K20017:
What is the reason for refreshing the radio page?

cwburnett:
> What is the reason for refreshing the radio page?

If the radio is updated by the ATC window for example, or another tool like vpilot, we want that change to be reflected on the FMS in realtime

cwburnett (approved):

cwburnett (approved):

PR #103, opened 2020-09-26
add support for data storage and file reading into a new core library

kaosfere:
This moves my config loader code into a new common library location and extracts's @tavip's data storage interface from Common_PFD_MFD.js.

I also renamed classes with a 'WT' prefix as we had discussed doing for 'library' functions to protect namespaces.


dga711:
@kaosfere I don't think we should not call it DataStorage.js  ? Even if stuff is case sensitive. Maybe call it like the class?

kaosfere:
> 
> 
> @kaosfere I don't think we should call it DataStorage.js ? Even if stuff is case sensitive. Maybe name it after the class?

Fixed!

dga711 (approved):
good stuff!

PR #114, opened 2020-09-27
move custom engine displays into a new class with their own config file

kaosfere:
this will allow us (or others) to create configurations for the new XML engine gauges without needing to stomp on the stock panel.xml.


mattnischan (commented):
What are the new xml engine gauges you're referring to?

kaosfere:
> 
> 
> What are the new xml engine gauges you're referring to?

@knighty has coded support for several new gauge types as well as multiple engine pages on the MFD.     It requires additional XML configuration, defining several EngineDisplayPage items, each of which can contain gauge elements.    The original implementation meant needing to modify panel.xml to configure it.  This this change you can leave panel.xml alone, and the WTEngine looks for its own config file, using that if its there otherwise defaulting to what's in the stock panel config so it fails back gracefully.

kaosfere:
Compare the WTEnginePanel.xml in this to the full modified bonanza panel.xml here:

https://github.com/Working-Title-MSFS-Mods/fspackages/blob/77368a5d557f6e835b689090f29579665f3f6e2f/src/workingtitle-aircraft-bonanza-g36/SimObjects/Airplanes/Asobo_Bonanza_G36/panel/panel.xml

And the stock one.

knighty (approved):
As discussed on Discord, I approve of this change. Will make compatibility with other mods that change panel files much better 👍

PR #118, opened 2020-09-27
PFD and MFD wind bug fixes

kaosfere:
1) move the AS1000 MFD wind code into CommonPFD_MFD like the core
files did in 1.8.
2) only display NO WIND DATA on PFD when on the ground
3) when wind < 1mph:
 a) PFD shows 0mph and 0deg
 d) MFD shows no wind arrow
4) slight performance increase by shortcutting when on the ground


knighty (approved):
Looks fine to me!

PR #123, opened 2020-09-28
changes!

kaosfere:
Cleaned up some redundant links in HTML
... release notes!


knighty (approved):
I have a terrible memory but this looks like it covers most of the stuff I personally did, without going into excessive detail.

PR #124, opened 2020-09-28
added stub for experimental UI work

kaosfere:
This is the section I had talked about creating for the work that @tavip started in his RFC #100.


kaosfere:
Going to close this one out since this has been handled by an Asobo patch.   If we want to do more UI work we can revisit then.

PR #129, opened 2020-09-28
Always apply the first successfully loaded engine page

knighty:


src/workingtitle-vcockpits-instruments-navsystems/html_ui/Pages/VCockpit/Instruments/NavSystems/AS1000/WorkingTitle/Engine.js
```diff
@@ -49,8 +49,8 @@ class WTEngine extends Engine {
                         });
                     }
                     this.engineDisplayPages[id] = engineDisplayPage;
-                    if (this.selectedEnginePage == null) {
-                        this.selectedEnginePage = id;
+                    if (i == 0) {
+                        this.selectEnginePage(id);
```
kaosfere, line 53: Isn't this going to break if the default page isn't the first one in the panel xml?

kaosfere: Discussed on discord.   This is fine, it makes the assumption that the "main" page will be the first one in the DisplayPages section, but that would have probably been the case anyway.

kaosfere (approved):

PR #143, opened 2020-09-29
Fix automatic building of package on Release

mattnischan:


dga711:
Screwed by rebase. See #147

PR #158, opened 2020-10-01
putting WT stuff into its own home

kaosfere:
This is done with the intention of furthering the separation we've been talking about, with the express goal of solving the problem of us breaking the G3000 synthetic vision with the latest update.


knighty (approved):
Definitely approve of this unsurprisingly based on my comments in the directory restructure issue. Unfortunately as we've discovered it's not easy whatever way we do it to solve the issue of files which get included via other ones that we don't modify, but given the choice between the 2 schemes this is definitely the best imo.

PR #159, opened 2020-10-01
break up vcockpits-instruments-navsystems into separate sysem packages for individual development

kaosfere:
After poking around with a number of different layouts, and thinking about how we can most safely develop different avionics systems that all "live" in asobo-vockpits-instruments-navsystems, I set this up.  The true "shared" code lives in the main package, each individual system is now in its own package with everything under WorkingTitle/ except for a stub html file in the stock location that just includes our stuff like we discussed this evening.


kaosfere:
Oops.   Didn't mean to push 3d50940 to this branch, but I'm going to leave it.  That's kinda the whole reason I'm doing this in the first place.

knighty (dismissed):

PR #164, opened 2020-10-01
CJ4: Override sim fuel consumption model

mattnischan:
The code here is ready for review, but it would probably be good to have some other folks take some flights and make sure they're getting reasonable numbers.

++ fuelmgmt page refactoring
++ navradio page refactoring


dga711 (approved):
code-wise i approve

dga711 (approved):

PR #202, opened 2020-10-07
CJ4: dep/arr page fixes

tscharlii:
Arrivals: Enable paging for transitions, e.g. EDDK: NVO transition is selectable now
Arrival/Approach Filter: trailing space in approach runway does not prevent match, e.g. EDDC
Minor improvements to arrivals and departure pages

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_DepArrPage.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_DepArrPage.js
@@ -167,7 +167,7 @@ class CJ4_FMC_DepArrPage {
             }
         }
         let rowsCount = Math.max(displayableRunwaysCount, displayableDeparturesCount);
-        let pageCount = Math.floor(rowsCount / 5) + 1;
+        let pageCount = Math.max(Math.ceil(rowsCount / 5), 1);
 
 
         //start of CWB EXEC handling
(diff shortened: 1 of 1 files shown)


knighty (commented):
I don't do work on the CJ4 but tabs need changing to spaces throughout in several places

dga711:
@tscharlii Thank you for your contribution. I will test it later today, as this one was on our todo list.

dga711 (changes requested):
Did test it for a little bit. Here are my findings. Which must not be all caused of your changes I didn't really do back and forth testing. But it would be nice to have them fixed.

- https://streamable.com/p819tq It scrolls away the runways, which i think should stay in that case. But we would have to ask @K20017 what the correct behavior is
- same as above for approaches and trans when an approach is selected
- https://i.imgur.com/n0fwEim.png the stars and approaches should be small text. add [s-text]

For a better reference, here is part of a video @K20017 made using the rw FMS https://streamable.com/ms41n4

In general our display of the ARR page is not yet 100% true to real life. But as a first iteration I think it would be good enough with the points fixed.

tscharlii:
Working on it. The first and second point require more complicated changes than the commits so far and I did not get it right on my first try. So I have some questions in return:

- Is there another way to test the logic than starting the game and hopping in a CJ4? The last picture on the announcement page https://forums.flightsimulator.com/t/cj4-community-mod-by-working-title-released/280375, which shows the about page, looks like some kind of simulator for the simulator (inception detected)
- How do I view the javascript console and see exceptions, console.log output, maybe syntax errors

dga711:
> * Is there another way to test the logic than starting the game and hopping in a CJ4?
You already know about this ? https://github.com/dga711/msfs-webui-devkit

tscharlii:
Didn't know it yet. That sounds about right.

dga711 (approved):
@tscharlii That was quick. I tested it and it was flawless. Really really good!

Again, thank you for your contribution. This helps a lot!

tscharlii:
You are welcome. I enjoy flying with this mod very much! Glad to help improving it.

PR #206, opened 2020-10-07
Features/g3000 rangecompass

cywang90:
Main features are implementing a range ring and range compass (and supporting code). Other G3000-specific changes related to this.

Importantly, there are also major additions/changes to MapInstrument.js, one of which will break backward compatibility: MapInstrument will no longer keep track of "_ up" orientation modes. Calling rotateWithPlane() does nothing now. Other methods/variables related to keeping track of orientation mode were removed completely. Map rotation is now handled by setting mapInstrument.rotationCallback. The callback will be polled every update step to get the desired map rotation, and MapInstrument will handle everything else.


src/workingtitle-base-aircraft-common/ModelBehaviorDefs/Asobo/GlassCockpit/AS3000.xml
```diff
@@ -169,7 +169,6 @@
 			<EMISSIVE_CODE>(L:XMLVAR_AS3000_DisplayLighting, Percent over 100) #LIGHT_DIMMING_SCALE# * #MAX_EMISSIVE_PERCENT# #MIN_EMISSIVE_PERCENT# - * #MIN_EMISSIVE_PERCENT# +</EMISSIVE_CODE>
 		</UseTemplate>
 		<UseTemplate Name="ASOBO_AS3000_Screen_Template">
-			<SCREEN_EMISSIVE_CODE>0.5</SCREEN_EMISSIVE_CODE>
```
kaosfere, line 172: Was this removal intentional?

src/workingtitle-vcockpits-instruments-navsystems-g3000/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/AS3000/AS3000_Common.js
```diff
@@ -157,22 +157,51 @@ class AS3000_MapElement extends MapInstrumentElement {
 			}
 		}
 		
-		if (this.lastOrientation != orientation) {
-			switch (orientation % 3) {
+		if (this.orientation != orientation) {
+			switch (orientation) {
 			case 0:
```
kaosfere, line 162: I still think it might be useful to add constants in MapInstrument for the various orientations.   Feels on that?

src/workingtitle-vcockpits-instruments/html_ui/Pages/VCockpit/Instruments/Shared/Map/WorkingTitle/MapInstrument.js
```diff
@@ -1215,44 +1217,18 @@ class MapInstrument extends ISvgMapRootElement {
...
-		} else {
-			// MOD: if orientation is heading or track up, we want to place the plane 33% from the bottom of the map,
-			// but vector needs to be adjusted for overdraw factor of sqrt(2), which is where the magic number of 382 comes from (500-500/sqrt(2)+0.33*1000/sqrt(2))
-			let target = this.navMap.XYToCoordinatesFromPlaneWithRotation(new Vec2(500, 382));
-			this.setNavMapCenter(target);
-		}
+		// MOD: allow for arbitrary placement of plane
+		let r_x = this.navMap.aspectRatio < 1 ? this.navMap.aspectRatio : 1;
+		let r_y = this.navMap.aspectRatio > 1 ? 1 / this.navMap.aspectRatio : 1;
+		this.setNavMapCenter(this.navMap.XYToCoordinatesFromPlaneWithRotation(new Vec2(500 * (1 - r_x) + 1000 * (1 - this.planeTrackedPosX) * r_x, 500 * (1 - r_y) + 1000 * (1 - this.planeTrackedPosY) * r_y)));
```
kaosfere, line 1223: Could you do me a small favor and spread this across a couple lines to it's easier to read?   Sorry. :}

src/workingtitle-vcockpits-instruments/html_ui/Pages/VCockpit/Instruments/Shared/Map/WorkingTitle/MapInstrument.js
```diff
@@ -1215,44 +1217,18 @@ class MapInstrument extends ISvgMapRootElement {
...
     }
 	
     centerOnActiveWaypoint(_val) {
         this.bEnableCenterOnFplnWaypoint = _val;
     }
 	
-    // MOD: adapt this method to the new orientation model for compatibility purposes
+    // MOD: this method now does nothing
 	rotateWithPlane(_val) {
```
kaosfere, line 1231: What about a change in the default rotation handler so that it can return either 0 or HDG based upon the status of a flag set within it, but defaulting to 0.   Then we could update `rotateWithPlane` to toggle that flag in the default handler.   This would, if I'm thinking clearly, still let us keep backwards compatibility.   But I might be missing something.

src/workingtitle-vcockpits-instruments-navsystems-g3000/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/AS3000/AS3000_Common.js
```diff
@@ -3,7 +3,7 @@ class AS3000_MapElement extends MapInstrumentElement {
         super();
 		this.simVarNameID = _simVarNameID;
 		
-		this.lastOrientation = 0;
+		this.orientation = 0;
```
kaosfere, line 6: I like this overall, mainly have that thought on compatibility in MapInstrument.js.    If we can get that figured out I'd be in favor of merging this, although I would suggest perhaps we integrate the G1000 and GX with this branch first if whe don't keep compatiblity in MapInstrument.

kaosfere:
I like this overall, mainly have that thought on compatibility in MapInstrument.js. If we can get that figured out I'd be in favor of merging this, although I would suggest perhaps we integrate the G1000 and GX with this branch first if whe don't keep compatiblity in MapInstrument.

src/workingtitle-base-aircraft-common/ModelBehaviorDefs/Asobo/GlassCockpit/AS3000.xml
cywang90: Yes. That was a regression (have no idea how that code got back in there). If that line is left in it will set a static screen brightness.

src/workingtitle-vcockpits-instruments-navsystems-g3000/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/AS3000/AS3000_Common.js
cywang90: If we want MapInstrument to define the constants, it would make most sense for MapInstrument to define rotation behavior as well (i.e. how the code used to work). Otherwise we have the possibility of outside code using the same constants but then assigning completely different rotation behavior to them, which seems messy to me. I'm not entirely opposed to having MapInstrument handle the orientation definitions again. I'm curious what other people think. Should all the orientation handling live within MapInstrument?

src/workingtitle-vcockpits-instruments/html_ui/Pages/VCockpit/Instruments/Shared/Map/WorkingTitle/MapInstrument.js
cywang90: Yeah, I can definitely do something like this. How I do it will depend on whether we want to move all the orientation handling code back into MapInstrument per https://github.com/Working-Title-MSFS-Mods/fspackages/pull/206#discussion_r502108316

src/workingtitle-vcockpits-instruments-navsystems-g3000/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/AS3000/AS3000_Common.js
kaosfere: No, I think you're mostly right... tere's really no worry about individual logic using its own values if it's the only thing that's dealing with rotation.

The *one* thing that I'm slightly hung up on would be communicating status via LVar, since that can only be bool/number.  If we wanted to share the orientation mode across instruments for some reason we'd have to have an agreement on that the number represents.

But I might be overthinking / prematurely optimizing this.  I tend to do that sometimes.

src/workingtitle-vcockpits-instruments/html_ui/Pages/VCockpit/Instruments/Shared/Map/WorkingTitle/MapInstrument.js
kaosfere: After I left that comment I stepped away from the computer and realized that since we're using our own instrument backwards compatibility is a limited concern.  The only things using that are the 1000, 3000, and G3X, and we can easily adapt all those in one go.  Doing extra work for compatibility that we can easily never need to use doesn't make sense...   so I think I'm fine with leaving this as is.

src/workingtitle-vcockpits-instruments/html_ui/Pages/VCockpit/Instruments/Shared/Map/WorkingTitle/MapInstrument.js
cywang90: Done. :)  e41ab0d8cb879829c1b3ee725b1e388c03090a44

kaosfere (approved):
Looks awesome to me.

PR #208, opened 2020-10-08
Add Wwise project

J-Hoskin:
Adds source Wwise project so we can create amazing sounds


dga711:
I will pull it into a working branch in our repo, so we can have a better look at it. Thanks a lot for doing it.

PR #209, opened 2020-10-08
Improve MFD UPPER MENU

J-Hoskin:
Improves the MFD Upper menu by:

- Redesigned FMS Info to resemble real menu and made working
- Added MFD checklist
- Added Passenger brief
- Added system overlay menus
- Remapped system button to new overlay
- Added functionality to PASS BRIEF CCP button
- Added functionality to CHK CCP button


dga711:
My findings on FMS Text.

- sometimes i get an exception in console that activewaypoint is not found. at those times there is no active waypoint but code should handle it
- I have setup a plan from EDDP to EDDK with a departure procedure. Blue is empty while still on ground. Think it should show Origin or Runway when departure is set
image
But it 
- afaik distances > 100nm should have no decimals

J-Hoskin:
@dga711 Thanks for your testing.

1.  I will fix this tomorrow
2. Blue is only ever the waypoint you last passed. If you haven't taken off an gone past a waypoint, then it remains blank.
3. Good observation, I'll implement that

K20017:
When on the ground, the blue waypoint should show the departure airport (when one has inputted it on the FPLN page) and then it will change to the runway (RW09L for example) once you have loaded a runway in on the DEP/ARR page.

dga711:
Replaced by #211

PR #216, opened 2020-10-09
buildps1: remove purge argument

dga711:
Earlier I added the PURGE argument to robocopy to remove files in the output that are not in source anymore.

But it turned out to work badly on packages with multiple assetgroups, as the former would always get deleted. So I removed it for now.


kaosfere (approved):
APPROVED!  HAPPY?  :D

PR #229, opened 2020-10-11
add data storage backing to all G3000 TSC settings 

kaosfere:
... and to the PFD_Attitude in CommonPFD_MFD while we're at it.


cywang90 (approved):
Looks good to me!

PR #258, opened 2020-10-14
CJ4: FMC Waypointloader performance

tscharlii:
The main and general idea of this PR is: **Improve flight plan editing performance** and make it a more fluent experience.

Do that by loading less data from the waypoint database. Consider and focus on the task at hand and load what's needed for that. Load additional data later, if at all.

Programming a flight plan using the PR changes: https://www.youtube.com/watch?v=zFo73ZLyZOg
Programming the same flight plan using the current v0.4.1 release: https://www.youtube.com/watch?v=oRoe-TV1NDY

Flightplan: EDIW DEXEN Y124 AMPIT L15 HON L10 LAM L610 NIK L179 ROMIN EDDL
From Dublin over to Wales, going SE and pass London near Heathrow, finally going E through Belgium to Düsseldorf.

The PR also adresses two issues:
- From known issues: Sometimes when entering an airway and the exit fix, the airway name will change to the shared airway, however the routing is still correct.
- Some VORs don't want to be part of an airway (see bogus DIRECT entries in flight plan in the 0.4.1 video)

### Technical details
**Departure and Destination Airports**
Status Quo: All approach fixes belonging to the airport will be loaded.
Effect: "Working..." until done.
PR Change: Do not load all the approach fixes, when selecting an airport. The departure airport does not need approach fixes at all, and the destination airport approach fixes can be loaded when we select a specific arrival/transition/approach
Example: See selection of destination airport EDDL

**Airways/Waypoints**
Status Quo:
1. All the waypoints from the airway's entry to exit node will be loaded.
2. All the waypoints from any other airway that touches/crosses our currently selected airway are also loaded. Most of them are not needed. We will probably continue on another airway from our exit fix. So we will only need one airway there. We dont know which one yet. We won't need fixes from crossing airways of intermediate fixes at all.
Effect: "Working..." until done

PR Change:
Still load the fixes on the airway (1.). They end up on the flightplan.
Defer loading of all the other fixes, that are directly connected to the current airway, to a background task (not showing "Working..." anymore). Some of them might be useful later.

Two things can happen:
1. The airway does not have many adjacent airways. Background Loading will be fast and is done by the time we enter the next airway. We can enter it quickly and do not get waiting times. See Y124 from DEXEN to AMPIT and L15 from AMPIT to HON
2. There are many fixes to load and the next intended airway is not loaded yet, when we enter it: The FMC then specifically requests this airway/these fixes now and we get a short "Working..." time. See entering airways L10 from HON to LAM and L610 from LAM to NIK in the video.

**Correct Airway designation**
Each waypoint now holds the identifier of its user-selected incoming and outgoing airway. Can be undefined.
Do not rely on Intersection.getCommonAirways anymore, which could yield the wrong shared airway.
For instance from DEXEN to AMPIT there are 2 airways: Y124 and UY124, the lower and upper variants. getCommonAirways would return one of those, UY124 in the example case although we selected Y124.


cwburnett:
@tscharlii thanks for this, we will review today! Looks promising!

dga711:
@tscharlii Thanks for the PR. And thank you for diving so deep into this topic.
We also had some investigation into it but didn't follow up on it yet.

On my first run, performance was very impressive. Right now I am trying to break it and here is my first find.

It seems like it  has some problems when an approach is loaded. The FPLN won't take any new waypoints. And on the second (false) Exec it will spit out the Exception you can see at the end of the video.
https://streamable.com/gotsfx

P.S: Do you have Discord? If you would be interested in working more closely with us on the mod, we would like to invite you to the core dev team. We think you would be a good addition :-)

mattnischan:
@tscharlii - This overall looks quite nice: I had actually gone in this direction for 0.4.0, but due to the issues you're addressing with airways I pulled it in favor of just having most waypoints load from the new loadFacilityRaw method for now until I could finish the implementation. But you beat me to the punch!

Can you annotate a bit here with some comments the changes you've made in FlightPlanManager.js and Waypoint.js?

tscharlii:
Sure, I can elaborate more on the changes tomorrow. In the mean time: You did notice, that a93f749e38c0d148ad05b057ff9ef0b28df469a4 shows the diff of these two files with respect to the original versions from MSFS 1.9.5, didn't you?

mattnischan:
Oh nice! I had actually not seen that you separated the commits between the addition and the changes, thanks a bunch for that.

I'll definitely take a look.

tscharlii:
> P.S: Do you have Discord? If you would be interested in working more closely with us on the mod, we would like to invite you to the core dev team. We think you would be a good addition :-)
tscharlii#9874

dga711 (commented):
Code-wise it looks absolutely fine to me.
My testing also went well, besides the problems when an approach is set.
👍

mattnischan (approved):

dga711 (approved):

cwburnett (approved):

PR #266, opened 2020-10-15
CJ4: TO/App Refs VSpeed send flow

dga711:
Contains

- send flow for vspeeds
- extracted ApproachRef to its own class and file
- fixes some formatting errors on the pages

Known issue:

- When sending is in progress and you navigate away, it will bring you back to the page after it is sent. Can't think of a good way around that.

DEMO: https://streamable.com/qvs9nl


mattnischan (approved):

K20017:
Wow nice.

PR #280, opened 2020-10-16
CJ4: SimBrief import

dga711:
Import current SimBrief flight plan into the FMS
https://streamable.com/ky7m5a

**TODO**
* what happens when waypoint is not found? right now it just skips it
* [x] set pilotid in "Mod Settings" @cwburnett will do it
* try setting DEP and ARR? i think its not useful as the data differs too much
* [x] set flight number 
* [x] set cruise level
* could set pax number, but would need to trigger weight calc
* could set weights from plan, but dunno if a good thing
* any other useful data from flight plan? perf init things?
* ...


cwburnett:
@dga711 could use your review on what I did in mod settings page. Not sure if I got this even remotely right ;-)

dga711:
looks super good @cwburnett 👍

cwburnett (approved):

PR #295, opened 2020-10-17
Develop/mapelement refactor

cywang90:
Pulled a lot of code from AS3000_MapElement.js into WT_MapElement.js, which lives in the shared navsystems folder. Also transitioned the system completely to use datastore instead of SimVars, which simplifies things since datastore is not async.

### DIFF
--- a/src/workingtitle-vcockpits-instruments-navsystems/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/Shared/WT_MapElement.js
+++ b/src/workingtitle-vcockpits-instruments-navsystems/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/Shared/WT_MapElement.js
@@ -1,0 +1,395 @@
+class WT_MapElement extends MapInstrumentElement {
+    // _settings should be an iterable of WT_MapSetting or WT_MapSettingGroup objects
+    constructor(_simVarNameID, _settings = [], _syncVarName = WT_MapElement.VARNAME_SYNC, _syncInitVarName = WT_MapElement.VARNAME_SYNC_INIT_DEFAULT) {
+        super();
+        this.simVarNameID = _simVarNameID;
+        this.syncVarName = _syncVarName;
+        this.syncInitVarName = _syncInitVarName;
+        
+        this._settings = Array.from(_settings);
+        this._settingsToSync = [];
+        for (let setting of this._settings) {
+            if (setting.toSync) {
+                this._settingsToSync.push(setting);
+            }
+        }
+        
+        this._lastSync = 0;
+    }
+    
+    addSetting(_setting) {
+        this._settings.push(_setting);
+        if (_setting.toSync) {
+            this._settingsToSync.push(_setting);
+        }
+    }
+    
+    callSettingsOnTemplateLoaded() {
+        for (let setting of this._settings) {
+            setting.onTemplateLoaded();
+        }
+    }
+    
+    onUpdate(_deltaTime) {
+        super.onUpdate(_deltaTime);
+        
+        let sync = SimVar.GetSimVarValue(this.syncVarName, "number");
+        if (sync != this._lastSync) {
+            if (sync == WT_MapElement.Sync.ALL) {
+                let initIndex = SimVar.GetSimVarValue(this.syncInitVarName, "number");
+                if (initIndex >= 0) {
+                    let initID = WT_MapElement.SYNC_INITID_ARRAY[initIndex];
+                    if (initID == this.simVarNameID) {
+                        this.syncMasterToAllSettings();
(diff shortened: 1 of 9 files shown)


src/workingtitle-vcockpits-instruments-navsystems/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/Shared/WT_MapElement.js
```diff
@@ -0,0 +1,384 @@
+class WT_MapElement extends MapInstrumentElement {
```
kaosfere, line 1: I believe we decided when we set up our "shadow" filesystem that entirely new files that went into /WorkingTitle/Pages/ didn't need to have a prefix on them.   It's important for class names in code because everything gets glommed into one namespace, but since we load the js with explicit path names it's a bit redundant  having /WorkingTitle/Pages/WT_something.js.

kaosfere: For classes, I think the cases where we have new ones we've not used an underscore separating the prefix (eg, WTDataStore).   But I think @cwburnett  may have said that he does?   I'm not sure we have a consistent style on this, come to thing of it.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/Shared/WT_MapElement.js
```diff
@@ -0,0 +1,384 @@
...
+        super.onUpdate(_deltaTime);
+        
+        let initID = WT_MapElement.getSettingVar(WT_MapElement.VARNAME_SYNC_INIT_ROOT_DEFAULT, WT_MapElement.VARNAME_SYNC_ALL_ID, "");
+        if (initID == this.varNameID) {
+            this.syncMasterToAllSettings();
+            WT_MapElement.setSettingVar(WT_MapElement.VARNAME_SYNC_INIT_ROOT_DEFAULT, WT_MapElement.VARNAME_SYNC_ALL_ID, "");
+            WT_MapElement.setSettingVar(WT_MapElement.VARNAME_SYNC_ROOT, WT_MapElement.VARNAME_SYNC_ALL_ID, WT_MapElement.Sync.ALL);
+        }
+        
+        let sync = WT_MapElement.getSettingVar(WT_MapElement.VARNAME_SYNC_ROOT, WT_MapElement.VARNAME_SYNC_ALL_ID);
```
kaosfere, line 39: I feel like there's some inefficiency here in (possibly) setting the value in the prior if statement then fetching it in this one.   It's small, but it makes me flinch a little.   If `(initID == this.varNameID)` you already know what `WT_MapElement.getSettingVar(WT_MapElement.VARNAME_SYNC_ROOT, WT_MapElement.VARNAME_SYNC_ALL_ID)` is going to be.  Maybe make that a trinary assignment based on whether  `(initID == this.varNameID)`?

kaosfere (changes requested):
This looks fairly good.    Two general comments:

First, it took me a while to make sense of what was going on  in WT_MapElement, and it would have taken me even longer if I hadn't already worked through the synced-settings logic you had in the G3000.     I think WT_MapElement.js could do with some documentation via jsdoc to make it a bit easier to grasp what everything is when you're working with it.

(@mattnischan  has been pushing for us to document anything "all new" with jsdoc anyway, and I like the idea, too.)

Second, I think there's a lot of logic here that's not map related and probably doesn't really belong in subclass of MapElement, but rather in an entirely new library that can be used by multiple instruments for setting/syncing/preserving settings.

Pulling this out into that, though, is probably a bit of premature optimization so I'm not going to ask for that at this point.  It's something to keep in mind, though.

src/workingtitle-vcockpits-instruments-navsystems/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/Shared/WT_MapElement.js
cywang90: resolved with a8b9064

src/workingtitle-vcockpits-instruments-navsystems/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/Shared/WT_MapElement.js
cywang90: resolved with 8673401

kaosfere (approved):
Awesome!   Thanks for the documentation in particular.  😄

PR #310, opened 2020-10-18
CJ4: Add frequencies page

tscharlii:
Initial version of a frequency page

TODO
- [x] MULTIPLE entries cannot be opened yet
- [x] Airports without frequencies show Pagecount 1/0
- [x] Page crashes for good, when entering a pilot defined waypoint, that is not an airport.
- [x] remove from CJ4_FMC_InitRefIndexPage and put it into a separate class file
- [ ] ~~Show frequency of VOR, NDB, ILS, Localizer, if entered as pilot-defined or Origin/Destination??? @K20017: Is it possible to view non-airport waypoints on this page? I know, you can start the sim with arbitrary Waypoints set as origin or destination from the MSFS main menu.~~

ISSUES
- [x] Fix blank screen on long frequency name
- [ ] Find a way to shorten the most common frequency names while keeping relevant information. The names are long in recent Navigraph Data, e.g. "ILS/GS CAT III RW25C"
- [x] Frequencies are off by 0.005 MHz, eg. one EDDB Ground frequency is 129.505, but the frequency page says 129.510


tscharlii:
Frequency Page Usage:
Top LSK: Cycle through configured airports: Origin/Destination/Alternate(NYI, always empty)/pilot-defined airport
Top RSK: Set and select pilot-defined Airport entered via scratchpad
LSK or RSK next to a frequency: Copy to scratchpad (only if scratchpad is empty)
MULTIPLE: there are more frequencies of a specific type (e.g. "Ground") at that airport: Click to view them

cwburnett:
@tscharlii I am getting this black screen when trying LSK L1 https://youtu.be/w5Y73v4_j4U

dga711 (approved):

PR #314, opened 2020-10-18
CJ4: Fix direct to approach fixes

dga711:
Did some minor tweaks. It seems to work better now.
https://streamable.com/mmjwz9

Please test test test


K20017 (approved):
Good

Issue #296, opened 2020-10-17
CJ4: v0.5.0 FMS Text Distance Problem (still)


cwburnett:
See video.
[https://youtu.be/5pfZCP3QGEE](https://youtu.be/5pfZCP3QGEE)

dga711:
![Uploading image.png…]()

PR #318, opened 2020-10-18
CJ4: Fix calculation bugs in the FMS Flight Log Page

J-Hoskin:
ETE and Landing time would continually update and never stop.


mattnischan (approved):

cwburnett:
@J-Hoskin can we also look at this with this PR https://github.com/Working-Title-MSFS-Mods/fspackages/issues/296

PR #322, opened 2020-10-18
LEGS: k2's crazy dirto approach fix

dga711:


K20017 (approved):
Good.  This would have taken weeks to find

PR #341, opened 2020-10-20
Update CJ4_MFD.js Fix Elevator Trim Indicators

Danice737:
Fix Elevator Trim Indicators so they move the right direction along with a change to CJ4_SHARED.js

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/MFD/CJ4_MFD.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/MFD/CJ4_MFD.js
@@ -1167,7 +1167,7 @@ class CJ4_SystemOverlayContainer extends NavSystemElementContainer {
                 trimGroup.appendChild(this.AileronRadians);
 
                 var text = document.createElementNS(Avionics.SVG.NS, "text");
-                text.textContent = "ND";
+                text.textContent = "NU";
                 text.setAttribute("x", (blockPosX + 6).toString());
                 text.setAttribute("y", blockPosY.toString());
                 text.setAttribute("fill", "white");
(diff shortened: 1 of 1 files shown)


J-Hoskin:
This is currently being worked on and should be fixed for upcoming release. Also a duplicate issue so I will close. Thanks for the suggestion.

J-Hoskin:
Apologies ahah, I am in the train and thought this was a issue. I will review once I get home.

J-Hoskin (approved):
Nice job. In the future, since these fixes are so closely related, I would put them in one PR so its easier to work on and ensure both gauges are consistent.

dga711:
Thank you for your contribution!

J-Hoskin:
No worries :) I may have made a mistake then.

Also you can see in the picture below, the cursor is in the middle of the green. I can only assume that this is the natural position (no elevator). Which means the cursor needs to start in the middle of the green block? @K20017 can you confirm how this dam thing is meant to work?

image

Danice737:
I had this question when trying to find how far to move the green box, I couldn't find a good pic of gauge I could confirm was just before takeoff for a starting cursor setting. So I only moved the green bar to match. and not the 0,0 position of cursor. It's not a hard change to move the cursor when we know the default neutral indication.

K20017:
I assume the bottom of the green is 0 deflection on the elevator. The reason why the green box goes more into the nose down position is because you don't want the nose to pop up prematurely on takeoff.

J-Hoskin:
@Danice737 after a very confusing conversation with the others I still have no idea what these trims are meant to be. Are you please able to revert my label changes and we'll stick with what you had for now until someone says otherwise?

Danice737:
My change makes the cursors move and indicate in the right direction so when you add  pitch down the cursor moves to the top and when you add pitch up trim the cursors move to the bottom of the gauge. and fixed the green band on the ECIS,

dga711:
We already have a brunch up to touch up some things. But it was not your changes that caused the confusion.
It is all fine for now.

PR #343, opened 2020-10-20
CJ4: Updated Aileron Trim Effectiveness for better LNAV turn response (reduces 'fishtail')

cwburnett:
This is a simply PR to adjust the aileron_trim_effectiveness to improve the roll response of the autopilot and, thus, improve the AP's ability to follow commands. This specifically improves intercepts and large angle turns. This also provides the AP more roll authority to reduce/eliminate the tendency to get 'stuck' at a 30+ degree angle of bank.

Please give it a test.

aileron_trim_effectiveness = 2


dga711 (approved):
I didn't really notice the improvement too much. But what do I no. AP still works

cwburnett:
This has been depricated.

PR #353, opened 2020-10-22
CJ4: Makes the Autopilot follow the FD better...

Danice737:
....and not hunt the TRK Line or ILS also when AP is engaged turns are not a sluggish. 4 test flights with marked improvement and no adverse effect I could see.


dga711:
@cwburnett should be tested together with #343

dga711:
While it is hard to see the improvement, because its hard to test, we ( or @cwburnett ) thinks the changes are good. So we will merge that.
Thanks again! ;) Good PRs!

dga711 (approved):

dga711:
So. We realized we screwed up the values during a merge of a pull request from someone else.
After realizing that and bringing those values back to (almost) default it is now fine again.

But thanks for looking into it

PR #354, opened 2020-10-22
CJ4: Fixes Battery Amps Display in ECIS and SYS pages. 

Danice737:
From what I can tell seems battery load is actually tracked in WATTs (comments in Systems,cfg under item power from OSBO) not amps, Appling the Watts to Amps conversion formula seems to display correct values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ src/sound/.cache
 *.wsettings
 src/sound/GeneratedSoundBanks
 src/sound/FlightSim.mediaid
+/.vs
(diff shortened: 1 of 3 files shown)


Danice737:
image

mattnischan:
So, I think what I'm wondering with this is what is actually correct. Did they just goof up and specify everything in watts in the cfg even though it was supposed to be amps? I would be surprised if Asobo changed out the standard FSX/ESP battery modeling system, because that would break the simvar unit conversions everywhere.

dga711 (approved):

dga711:
Thank you for your contribution!

cwburnett:
@Danice737 We went ahead and accepted this PR, but there's a lot that still needs to be done on electronics. The actual behavior is still way off, so we'll be addressing that later on, but for now having the right units is a good step. Thanks!

Danice737:
I just knew -500 shouldn't be displayed and started looking that's when I found the watts listed in the comments and tried the Watts to Amps conversion and at least it looks right now. I'm sure a lot of those watt draw values are off :)  I added a placeholder temp since I got tired of looking at dashes on the display :) I had some battery code for temperature (based on draw/charge and ambient temperature. but then I found out the CJ4 has a Lithium Battery and that completely invalidated my code so I'm working on a new code for that.

K20017:
The CJ4 actually has an option for lead acid, lithium, and NICAD.   We had a NICAD in ours which is the one that is most susceptible to thermal runaway.
IMG_20200909_112736

Issue #250, opened 2020-10-13
CJ4: Mach mode climb


slip664:
When pressing the SPEED button to switch from IAS to MACH in FLC mode, the bug appears in the right spot on the airspeed strip but in the AP window, it just spits out a bunch of random numbers and doesn't climb at a MACH speed.

This may be something out of your control for now since I'm assuming it's tied into the MSFS autopilot.

![Capture](https://user-images.githubusercontent.com/72432870/95924035-a2dd3580-0d6b-11eb-93d1-d71734205d36.PNG)

dga711:
@K20017 Is this what you noticed yesterday, too?

zainsyedz:
I have the same issue, it does spit out random numbers, to provide more information, it climbs at the airspeed that matches the mach speed at the time of activation. For example let's say when setting the mach speed the airspeed was 250 knots, it climbs at 250 knots and doesn't maintain the mach speed selected.. Hope this clears it

Danice737:
#355 Fixes this issue

PR #355, opened 2020-10-22
CJ4: Fixes FLC Mach Climb display on AP Tape

Danice737:
Pulling wrong VAR was the cause and now works and display's correct value. Should fix issue #250


mattnischan (approved):

mattnischan:
@Danice737 Thanks for another contribution! Looks good to me.

PR #370, opened 2020-10-23
CJ4: Enhancement sets and tracks Cab Pressure values

Danice737:
So they can be tracked in the "NavBar" on the MFD. still needs code (which I haven't figured out how to do) to be able to switch from "Coms" to "Cabs" in the MFD (Menu or button?) Display code is already in the "CJ4_MFD.HTML" it's just Orphaned and also need this change to set the values to be displayed.

image

^^This Images is the orpahned code already written by Asobo. the file I changed calls and sets values to be displayed by this code already a part of the CJ4_MFD.HTML

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/Shared/CJ4_Shared.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/Shared/CJ4_Shared.js
@@ -3124,6 +3124,30 @@ class CJ4_NavBarContainer extends NavSystemElementContainer {
             else
                 this.ratElement.textContent = 0;
         }
+        // Add to track Diff Pressure 144 to convert PsF to PSI
+	if (this.diffElement) {
+            var prediff1 = Math.trunc(SimVar.GetSimVarValue("PRESSURIZATION PRESSURE DIFFERENTIAL", "Pounds per square foot"));
+            prediff1 = prediff1 / 144;
+            if (prediff1)
+                this.diffElement.textContent = prediff1.toFixed(1);
+            else
+                this.diffElement.textContent = "0";
+        }
+// add to track Cabin Altitude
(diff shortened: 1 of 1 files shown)


Danice737:
image

dga711:
Do you what the procedure would be to change the navbar? Is it a menu thing?

K20017:
Does it show anything for pressurization when you climb?  To be realistic, the pressurization bar should be on the captain's MFD but since we can't split screens, we opted to have the GS, TAS, RAT, SAT bar at the bottom in the mean time until we split them.  If pressurization doesn't work (which I don't think it does because you get pressurization errors on default), then we'd have to write up a pressurization controller to go along with the pressurization values.

Danice737:
Screenshot 2020-09-27 144601
This is what I was going by in the 14+ hours of IRL videos on the CJ4 I always saw it on the MFD....  

Yes It shows your Cabin Pressure and it rises as you climb and it shows the current rate and Diff pressure. 
the sim tracks all three in SimVar.
I just couldn't figure out how to get it to switch from Coms to cabs on the bar(I just hardcoded mine to always display the cabs.. I'm a C# programmer just now learning JS.

K20017:
> This is what I was going by in the 14+ hours of IRL videos on the CJ4 I always saw it on the MFD....
> Yes It shows your Cabin Pressure and it rises as you climb and it shows the current rate and Diff pressure.
> the sim tracks all three in SimVar.
> I just couldn't figure out how to get it to switch from Coms to cabs on the bar(I just hardcoded mine to always display the cabs.. I'm a C# programmer just now learning JS.

Interesting, I will have to test this then.  
Do you have discord?

Danice737:
yes I have discord

K20017:
Join us at: https://discord.gg/Fa6w2xK

cwburnett:
I think we need to hold until we can get the MFDs split, then we can have the CABS on one MFD and the existing NAVBAR on the other. Definitely on the to do list after we clean up performance a bit. Let's leave this open for the time being 👍

dga711:
The mod setting is actually a great idea

PR #378, opened 2020-10-23
CJ4: fix transponder toggle

panourgue:
On/standby was the other way around...
plus (probably unnecessary) code simplification


panourgue:
oops. already fixed :)))

Issue #309, opened 2020-10-18
CJ4: No Flashlight in Cold & Dark Cockpit


OliveBeast20441:
First off, thank you for continuing to develop and modify this aircraft. All of the changes so far have resulted in a much more stable flying model. The climb out to altitude with Version 0.5.0 was flawless, with no porpoising or other issues, other than a reluctance to switch from IAS to MACH mode and adjust pitch to remain on the mach number setting, but that may be a sim issue, out of your control.

However, one feature that existed with the base model was the availability of a "flashlight" (for lack of a better word) when in the cockpit at night time. You could at least turn on the batteries and avionic switches, as well as the overhead rotary knobs before starting the main engines. Now it's completely dark and I find myself fishing for a tool tip or other means to light up the interior.

But never the less, many thanks for bringing this aircraft up to snuff!

dga711:
@OliveBeast20441  To me it also doesn't turn on automatically right now. But I cannot think about somehing we do, that would cause this 🤔 
But you can turn on the flashlight via ALT + L hotkey. Or check the mapping in your control settings.

cwburnett:
This issue is tied to the cabin lights. We're adding the feature to save the cabin lights settings. In the upcoming patch, the flashlight only comes on automatically if there are no lights on in the aircraft. However, if the cabin lights are set to be on because they're saved that way, ALT-L will have to be pressed to turn on the flashlight. If they're set to off at the end of a flight and saved that way, then at the beginning of the next flight, the flashlight would automatically come on if dark. Hope that makes sense.

PR #403, opened 2020-10-24
CJ4: Persistent cabinlights setting

### LINKED ISSUE #389: Passengers Cabin lights always "on"
I noticed that from version 5.0 to version 6.0 it is no longer possible to manage the passenger cabin lights from the FMC, which always remain "on" each time the program is restarted. The problem seems to be an interference with the avionics switch  device. I changed the following lines   of the system.cfg,  by putting the parameter 10 instead 28, this seems to fix in part the issue:
 
 #cabin
lightdef.31 = Type: 13 # Index: 0 # LocalPosition: 6.5,0,3 # LocalRotation: 0,0,0 # EffectFile: LIGHT_CJ4_CabinOmni # PotentiometerIndex: 10
lightdef.32 = Type: 13 # Index: 0 # LocalPosition: 3.6,0,2.72 # LocalRotation: 0,0,0 # EffectFile: LIGHT_CJ4_CabinOmni # PotentiometerIndex: 10
lightdef.33 = Type: 13 # Index: 0 # LocalPosition: 0.5,0,3 # LocalRotation: 0,0,0 # EffectFile: LIGHT_CJ4_CabinOmni # PotentiometerIndex: 10
lightdef.34 = Type: 13 # Index: 0 # LocalPosition: -3,0,3 # LocalRotation: 0,0,0 # EffectFile: LIGHT_CJ4_CabinOmni # PotentiometerIndex: 10

However, this modification does not allow the management of the passenger cabin lights from the FMC in any case.
Any suggestions to fix it? Thx

dga711:
closes #389

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
@@ -16,9 +16,9 @@ class CJ4_FMC extends FMCMainDisplay {
         this.paxNumber = 0;
         this.cargoWeight = 0;
         this.basicOperatingWeight = 10280;
-    	this.grossWeight = 10280;
-		this.zFWActive = 0;
-		this.zFWPilotInput = 0;
+        this.grossWeight = 10280;
+        this.zFWActive = 0;
+        this.zFWPilotInput = 0;
         this.takeoffOat = "□□□";
         this.landingOat = "□□□";
         this.takeoffQnh = "□□.□□";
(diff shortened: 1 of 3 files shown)


cwburnett (approved):

cwburnett:
@dga711 is it worth defaulting these always to off to eliminate the "flashlight bug"?
https://github.com/Working-Title-MSFS-Mods/fspackages/issues/309
I think if we default to off, then we'd close 309. Up to you though, we can also just document it.

dga711:
Could be worth it, i don't really have a strong opinion here
Only downside could be that pilots newer realize there are any cabin lights :)

J-Hoskin:
There is a hotkey for cabin lights i believe. Can u hook them up to that? Makes them a lil more discoverable

K20017:
> There is a hotkey for cabin lights i believe. Can u hook them up to that? Makes them a lil more discoverable

Do you mean the small panel behind the copilot's head so they can turn on and off the cabin lights?  It's not even modeled remotely well enough :/

J-Hoskin:
@K20017 No I mean there is a keyboard shortcut to turn on and off cabin lights. I guess most people don't know about it but it would be worth looking into hooking it up to our cabin lights

dga711:
@J-Hoskin But isn't that meant for the cockpit cabin lights?

cwburnett:
If we want to tie these up to a keyboard or set them to default to not break the flashlight, we can do that in a separate PR.

PR #406, opened 2020-10-24
CJ4: Revamped (a little) error messages in the FMC

panourgue:
**updated** (see conversation below) 

~~[example of the problem](https://streamable.com/u0pn96), at least I think it's a problem :)))~~

~~Tested on several pages, didn't see any adverse behaviour, but you use this function in many places, so idk.~~

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_PosInitPage.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_PosInitPage.js
@@ -51,7 +51,7 @@ class CJ4_FMC_PosInitPageOne {
         };
         this._fmc.onLeftInput[2] = () => {
             let value = this._fmc.inOut;
-            this._fmc.inOut = "";
+            this._fmc.clearUserInput();
             if (this._fmc.tryUpdateRefAirport(value)) {
                 this._isDirty = true;
                 this.update(); // TODO hmm, i think invalidate() would be the more right thing to call but...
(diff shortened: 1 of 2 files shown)


panourgue:
meh

panourgue:
thought I saw a problem, but no

dga711:
I think I remember (from @K20017 s video where he shows input checks) that your input will remain on error.

If that is true, then the problem is more with weird things going on with it in general on error and clr/del etc. which must be streamlined. Because yeah, it is annoying.

panourgue:
@dga711  I'm pretty sure that there are two types of error/message notifications in proline:
- related to user input (like "invalid entry") shown actually inside the scratchpad
- line underneath the scratpad is for messages/warnings that are more to do with general state from the fms

> weird things going on with it in general on error and clr/del etc. which must be streamlined

I made some changes already.  👀 [check this out](https://streamable.com/zxu8by) 👀 
But I a bit reluctant to PR this because it replaces several base methods from "fmc" object in runtime and some of your code.

dga711:
that looks pretty cool.

i knew that behavior, but imo i like the errors beneath better. but ofc it is no "study level" :-D

good work. i think we should let @K20017 give his opinion on all this and then continue

cwburnett:
Yes, will leave this one for @K20017 to approve or ask for a change....

panourgue:
For clarification: this pr just straight up removes last user input when "invalid entry" occurs. Video link in [my comment](https://github.com/Working-Title-MSFS-Mods/fspackages/pull/406#issuecomment-716623472) does not represet it.

dga711:
image

But as I already said, besides that, there is something funky going on when you start typing again and maybe we can fix that. I just can't really describe it. @panourgue

dga711:
As you can see in this video from @K20017 he presses CLR and it wouldn't go to show CLR/DELETE, but instead showing last input and start removing letters from that.
https://streamable.com/d2elfu

dga711 (changes requested):
see above

panourgue:
@dga711 Looks similar to [what I what I tried at home](https://streamable.com/zxu8by). Maybe I update this pr so you could at least look at my idea?

dga711:
yeah, i mean if you can get it to work it would certainly give us "study level" points :D

panourgue:
@dga711 Tried to make it as intuitive as possible and similar to @K20017's video.

dga711:
Thank you very much. Will test.

dga711 (approved):
Very nice!

cwburnett:
Looks great, merging, thanks!

PR #407, opened 2020-10-24
CJ4: Only show radar bug when WX is on

dga711:


cwburnett (approved):

Danice737:
This works except weather bug shows when first loading into plane, turns off if you cycle TERR/WX/OFF, Also bug comes back on anytime you hit the PFD NAV button to switch nav modes

PR #412, opened 2020-10-25
CJ4 deparr direct entry

tscharlii:
The idea of this PR: When pressing DEPARR, go directly to the depature or the arrival page, depending on the current situation.

The logic is as follows:
- Origin set and airplane on the ground -> Departure page
- Origin set and airborne and no destination set and less than 50nm away from origin -> Departure page
- Origin set and airborne and destination set and (less than 50nm away from origin AND not halfway through to destination) -> Departure page
- Origin set and airborne and destination set and (more than 50nm away from origin OR halfway through to destination) -> Arrival page

Index page in all other situations, these are:
- No origin (don't care about destination) -> Index page
- Origin set and airborne and no destination and more than 50nm away from origin -> Index page

Press DEPARR twice to get to index page.

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
@@ -50,6 +50,8 @@ class CJ4_FMC extends FMCMainDisplay {
         this.selectedWaypoint = undefined;
         this.selectMode = CJ4_FMC_LegsPage.SELECT_MODE.NONE;
         SimVar.SetSimVarValue("TRANSPONDER STATE:1", "Enum", 1);
+        this.currentInput = undefined;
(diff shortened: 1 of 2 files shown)


tscharlii:
Press twice DEPARR to get to index page has a minor some.
You actually have to press it twice. So if you are somewhere else in the FMS, press DEPARR and get to arrivals or departures, then press DEPARR again, all is good: you get to the DEPARR Index
But if you already are on arrivals or departures, and did something there, i.e. pressed some other button: If you press DEPARR now, you'd expect to get to the index page, but you get to the situation evaluator first. You get to the index page on the second DEPARR key stroke.
Alternatively you can use the LSK L6 on departure or arrival to go directly to the index page.

As long as we do not have a fms menu navigation router which is aware of the current page, I don't see a good way to implement it and hope/think this is good enough.

cwburnett:
@K20017 can you review the behavior to ensure you approve before we merge? I like it, but want your experienced perspective

mattnischan (approved):
Looks good to me from a code perspective.

K20017:
The logic is correct for the Dep/Arr index.  Looks good.

PR #418, opened 2020-10-26
CJ4: Persistent stored MFD MEM function

cwburnett:
This PR will store and load MFD Presets from datastorage.


J-Hoskin (approved):
Awesome! Nice jobb.

J-Hoskin:
Ohh, you save the SystemOverlay menu state on mem buttons now. I don't know if that is what happens in reality since its an overlay. @K20017  thoughts?

K20017:
I don't think the SYS overlay is stored.  I've never actually tried in real life cause we never had one of those up.

J-Hoskin:
@cwburnett yeh i'd recommend getting rid of it, doesn't feel rid and its so easy to open anyway

cwburnett:
@J-Hoskin it was in two of your premade profiles so I included the value in the save, but we can certainly remove it, no problem, will do tomorrow

J-Hoskin:
@cwburnett ahh yeh I remember accidently adding it. Thanks :)

cwburnett:
okay, system overlay has been removed, this is ready for testing by @J-Hoskin @K20017 @dga711 @mattnischan

PR #424, opened 2020-10-26
CJ4: Enable using VOR mode when not in an approach

mattnischan:


cwburnett (approved):

Danice737:
Check CJ4_PFD.js line 161 looks like you missed a semicolon

PR #435, opened 2020-10-27
CJ4: Waypoint Select Page Improvements

### LINKED ISSUE #428: CJ4 Request - How to handle duplicate waypoints in FMS
I have no idea if the RW FMS is capable or if the WT team is capable of working on this with the tools at hand. When presented with duplicate waypoints we are asked to pick which one is needed and relevant. Some are obvious (airport vs. VOR) and they all have LAT/LONG values. But, some FMS's will sort by relative distance to present position, closest waypoint on top of the list. This is very convenient and would be a great feature if it can be done.

cwburnett:
only shows duplicates (exact ident matches)
sorts by distance
reformat display to be more reflective of reality

image

closes #428

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_SelectWptPage.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_SelectWptPage.js
@@ -14,7 +14,6 @@ class CJ4_FMC_SelectWptPage {
             [""],
             [""],
             [""],
-            [""],
             [""]
         ];
         //console.log("search ident: " + ident);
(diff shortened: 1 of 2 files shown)


src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_SelectWptPage.js
```diff
@@ -17,21 +17,54 @@ class CJ4_FMC_SelectWptPage {
...
-                    t = " AIRPORT";
+                    t = " AIRPORT ";
+                    region = w.infos.region;
                 }
-                rows[2 * i] = [w.ident + t];
-                rows[2 * i + 1] = [w.infos.coordinates.toDegreeString()];
+                rows[2 * i] = [w.ident.padEnd(5, " ") + t + freq + "[d-text]", region + "[d-text]"];
+                rows[2 * i + 1] = ["  " + w.infos.coordinates.toDegreeString() + "[d-text]"];
                 fmc.onLeftInput[i] = () => {
                     fmc.setMsg("WORKING...");
```
dga711, line 68: i wonder if we should get rid of setting working... here.
it should get set by the caller already an i am not sure if we can run into sticky working... problems here

dga711 (approved):
good stuff!

src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_SelectWptPage.js
cwburnett: removed

Danice737:
if (page < Math.floor(waypoints.length / 5)) { 
should be 
 if (page < Math.floor(waypoints.length / 3)) {

couldn't go back to the prev page and found this.

PR #449, opened 2020-10-29
CJ4: Hotfix arrival speeds calculation

panourgue:
@cwburnett @dga711 getApproachRunway() == undefined if user did not select an approach. You guard against this in other places, but not here. And they ask me why I hate javascript 😋


cwburnett:
Ha, we just broke this earlier today fixing something else. Nice catch. Thanks!

cwburnett (approved):

PR #453, opened 2020-10-29
CJ4 Scratchpad & Random fixes

dga711:
Makes the errors in scratchpad render independent of inOut. So we don't have to block LSK on error, which introduced other problems.
Some minor other fixes which i met on the way.

- [x] scratchpad error independent of inout
- [x] simbrief import "errors" are now msg
- [x] no lsk event on rwy on route page
- [x] fix decel error when no runway


dga711:
@panourgue review pls

panourgue:
@dga711 What's an LSK? Is it L0..L6 and R0..R6 button?

K20017:
> @dga711 What's an LSK? Is it L0..L6 and R0..R6 button?

LSK is Line select key. The 6 buttons on the left and right of the FMC display. So you can identify them like LSK 1R or LSK 6L

panourgue (changes requested):
Suggestion: instead of plugging previous scratchpad's value into lsk, just close error message and put that value back? User would have to click lsk two times, but imo it's more intuitive

J-Hoskin (approved):

PR #454, opened 2020-10-29
CJ4: Enhancement - Waypoint Select Page Improvements

Danice737:
Shifted R and L button press code to match new format


Danice737:
Before change you would have to press red button to get green to load, with change you press the green button to load green line
Screenshot 2020-10-29 081817

dga711 (approved):
Thank you again for your contribuitions. Good catch!

PR #496, opened 2020-11-01
G3000: Change window title of the map sync settings

lars-reimann:
Change the window title of the map sync settings from "Map Orientation" to "Map Sync Settings" as specified in https://static.garmin.com/pumac/190-02046-02_a.pdf on page 202.

I also removed some seemingly duplicate code.


cywang90 (approved):
LGTM.

PR #508, opened 2020-11-03
Add missing declarations so IDEs can resolve references and provide auto-completion

lars-reimann:
This pull request should not have any effect on the actual mod. It solely aims to improve the development experience with an IDE like Webstorm by providing TypeScript type definitions and JavaScript implementations for declarations that are currently not in the src/ folder. Here is an example how it now looks in Webstorm. Note that all references to Avionics.* are resolved:

The remaining unresolved references are either shortcomings of Webstorm due to the dynamic nature of JavaScript, which we can work around later by changing the code, or maybe in some cases bugs in the our/Asobo's code.

Some other use cases for the types/ folder:
* It acts as a list of files we still need to integrate.
* We can reference the implementation of Avionics, Simplane, SimVar etc.


dga711:
While I like this in general. I am somehow a bit on the fence of adding all these files 🤔 

Isnt there a way to generate type description js/ts from this?

lars-reimann:
I removed all files from the Asobo folder that were unrelated to the project or that were already moved to src/. But I can look into condensing this further using the tools you mentioned.

The reason I kept the folder structure intact for Asobo's part was so it was easy to see where files were originally located. This does not apply to the files that are directly in the types/ folder, so we do not lose information if these are merged into a single file.

The reason I kept the implementation where possible was so we can look up what a function actually does since there is no documentation, yet (at least none that I know of). It might also point us at possible performance improvements: For example, `Avionics.Dictionary` internally uses an array and does a linear search over it when setting/getting values. We could replace the array by a `Map` or maybe directly use `Map` in favor or `Avionics.Dictionary`.

Anyway, like I said initially I'm open for change requests.

lars-reimann:
I've merged a couple of files that were directly in the types/ folder and had a common purpose.

So, I guess the remaining questions are:
1. Do we want to keep the JavaScript files with implementation or just TypeScript definition files (also see my comment above)?
2. Do we want to maintain Asobo's file structure or condense them somehow (also see my comment above)?
3. How many files are OK to get this merged?

I also want to stress that none of these files end up in the built packages. They are just used while developing.

kaosfere:
This is an interesting one.  I really love the idea, as that's something that's been a massive pain to me.   It's a lot of new files to add, but they're fairly well contained, so that doesn't bother me all that much.   The main problem I have with this is the administrative overhead it's going to present for tracking Asobo changes, and for avoiding conflicts with our stuff.

With every update we're going to need to resync this with the Asobo files.   Additionally, as we add new files we'll need to remember to remove them here, and vice-versa.  

(I do think it would be useful to have the full text, rather than just prototypes;   that will make "jump to definition" so much more useful.)

How do you see what playing out over time?

mattnischan:
@lars-reimann 

I think my opinion here is that this is definitely too large a magnitude of files to add to the repository that aren't in active use, but moreso than that, we don't own any of those files. We've purposefully taken the steps to only include those files which we have modified, to stay in the best possible standing, both legally and relationship-wise.

Can we instead set this up so someone can do it easily locally instead? We do already have the `/ref` folder gitignored.

kaosfere:
Providing a utility script that would scan a list of official packages and copy everything that is not a file that we have a version of (with appropriate logic checking for WT paths) that could be run on demand to generate there for someone might be an alternate solution.

lars-reimann:
> The main problem I have with this is the administrative overhead it's going to present for tracking Asobo changes, and for avoiding conflicts with our stuff. With every update we're going to need to resync this with the Asobo files. Additionally, as we add new files we'll need to remember to remove them here, and vice-versa.

> I think my opinion here is that this is definitely too large a magnitude of files to add to the repository that aren't in active use. 

The guys from FlyByWire have [dedicated branches](https://github.com/flybywiresim/a32nx/tree/asobo-1.10.7) in their repository containing the latest version of the relevant official packages. That could be useful in general, because we could use diff tools to see what has changed, also for stuff already in `/src`.

And in any case, is it really such a big issue to keep this list up-to-date? That would depend on how often and severely the official packages are changed, you guys probably already have experience with this.

Removing the file from `/types` would have to be checked during a code review whenever new files are added. There are "only" 67 files from official packages left currently anywat. And even if we were to forget to remove a file from `/types`, the worst that would happen is that the IDE does not know what a reference should point to, which is the kind of the same situation we have now. The created mods would work regardless.

> Can we instead set this up so someone can do it easily locally instead? We do already have the /ref folder gitignored, which is where I keep copies of the Asobo files locally during development, so VS Code can follow them somewhat.

> Providing a utility script that would scan a list of official packages and copy everything that is not a file that we have a version of (with appropriate logic checking for WT paths) that could be run on demand to generate there for someone might be an alternate solution.

That could definitely be done but from what I've seen when I manually went over the files the locations of the files do not match 100% anymore. Example:

```
workingtitle-vcockpits-instruments-airliners\html_ui\Pages\VCockpit\Instruments\Airliners\Shared\WT\Waypoint.js

vs.

asobo-vcockpits-instruments\html_ui\Pages\VCockpit\Instruments\Shared\FlightElements\Waypoint.js
```

Some solutions:
* Just match files based on their name. However, this could prevent some files from being copied over which means some references do not get resolved.
* Match files based on their content. This is more work to implement but can be more accurate - until the corresponding files diverge.
* Have some configuration file (e. g. JSON) that contains the final parts of the paths to the files that should be included in `/types` or `/ref`. This is easy to implement and accurate but we have to keep this list updated (also see discussion above about the code review) and it would not work if files are moved by Asobo. On the other hand the latter behavior could actually be useful since it would alert us immediately that we probably have to change something in `/src`.

Another problem is things like the `Avionics` object. I was unable to find the actual JavaScript files containing this (maybe someone knows their locations), so I've used dga711's connector to the Chrome console and some `toString` calls to get to the classes and functions.

> We've purposefully taken the steps to only include those files which we have modified, to stay in the best possible standing, both legally and relationship-wise.

Sure, the legal issues are definitely relevant, particularly for the reverse engineering of `Avionics` etc. That said, as mentioned above, FlyByWire are hosting the code in their repository. Moreover, most of the code currently in this repository has been developed by Asobo. Just adding something to a file does not suddenly change the ownership of the existing code in the file (and some files like [this one](https://github.com/Working-Title-MSFS-Mods/fspackages/blob/main/src/workingtitle-vcockpits-instruments-navsystems/html_ui/WorkingTitle/Pages/VCockpit/Instruments/NavSystems/Shared/NavSystemTouch.js) are even completely unchanged).

In any case, I do agree that this is a question that should be clarified before this PR is integrated, so I've changed it to a draft. Maybe this is something that one of the core members could clarify directly with Asobo using the email address they provided in [the last Q&A](https://www.twitch.tv/videos/784582715?t=00h55m59s) (msfsim@microsoft.com).

mattnischan:
@lars-reimann - I very much appreciate your enthusiasm here, but unless we get a specific blessing from MS and/or Asobo, we're just not prepared to add all of their code to our repository (regardless of what any other team is doing). There are other totally usable local ways of both tying up the files for reference as well as for diffing. We should be treating their code like any other external dependency: you would not add React's code to your repository, for example. So, whatever solution we choose should operate in a similar way.

We are in the process of connecting with MS and if anything changes on the raw files front we can certainly revisit this issue from that angle. If there is a non-"add all the Asobo files" approach you'd like to pursue, we're definitely open to that.

lars-reimann:
I'll make new PR for the updated workflow.

PR #526, opened 2020-11-06
Added new Garmin fixes and pulled into CJ4 for dep isolation

mattnischan:


src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/MFD/CJ4_MFD.js
```diff
@@ -210,6 +210,14 @@ class CJ4_MFD extends BaseAirliners {
                 this.systemOverlay.show(false);
             }
 
+            const rangeSelectDisabled = WTDataStore.get('WT_CJ4_RANGE_SEL_DISABLED', 0);
```
mattnischan, line 213: Substantive change here. Default for the range sel is now enabled, and the code we had in the element itself is pulled up here.

src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/PFD/CJ4_PFD.js
```diff
@@ -168,6 +168,15 @@ class CJ4_PFD extends BaseAirliners {
             }
             this.mapOverlay.setRange(this.map.range);
         }
+
+        const rangeSelectDisabled = WTDataStore.get('WT_CJ4_RANGE_SEL_DISABLED', 0);
```
mattnischan, line 172: Range sel show/hide

src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_DsplMenuPage.js
```diff
@@ -75,8 +75,8 @@ class CJ4_FMC_DsplMenuPage {
         fmc.updateSideButtonActiveStatus();
     }
     static ShowPage2(fmc) {
-        let rngsel = WTDataStore.get("RANGE_SEL", 0);
-        let rngSelSwitch = (rngsel == 0) ? "" : "green";
+        let rngSelDisabled = WTDataStore.get("WT_CJ4_RANGE_SEL_DISABLED", 0);
```
mattnischan, line 78: Swapped this logic around so the default was on instead of off.

dga711:
@mattnischan to repeat my suggestion, could we make the default 1 ? Nobody knows how to activate it without asking but everyone wants to see it.

mattnischan:
Yep, already done here, take a look above. :-)

It's a little confusing because the datastore setting is now whether or not the range sel is _disabled_, as opposed to whether or not it is _enabled_. So, if nobody set the value yet it still works.

dga711:
> Yep, already done here, take a look above. :-)
> 
> It's a little confusing because the datastore setting is now whether or not the range sel is _disabled_, as opposed to whether or not it is _enabled_. So, if nobody set the value yet it still works.

oh ok, i only looked at the code snippets i saw on the front page :D

dga711 (approved):

PR #537, opened 2020-11-06
Develop/g1000 rearchitecture

knighty:
This is here as a placeholder for my re-architecture work


Flaxx-Rat:
When now using the default G1000NXI Bush-Trips with any a/c using this gauge(s) are almost impossible. When loading any of those flights about 5 minutes later the flight plan begins to randomize the legs (as it seems). This happens not only to the build-in map (below an example)
2022-12-19 05_48,05_AB
but as well to all g1000 screens as well (seconds later the following screen shot)
2022-12-19 05_48,22_AB
It doesn't matter if flying with or without autopilot and seems start even when the sim is paused by [ESC].
I'm using the Store-Version, everything is up-to-date, and it happens even with disabled add-ons (empty community folder)

PR #586, opened 2020-11-30
CJ4: Cosmetic changes

flymypi:
Adapted some PFD and MFD elements to feedback from K20017 and reference videos/images found online.
These include:

- Dash pattern for radar altitude ground in altitude band
- Geometry of arrow indicating active AP side and toggling its color based on AP status
- Sizing and Spacing of PFD/MFD bottombar values and their units
- Removed overflow of Airspeed numbers over the the black cursor
- Removed overflow of Attitude indicator on bank angle arc
- Removed overflow of Altimeter numbers over the the black cursor
- Adjusted FMC screen position to better line up with touchkeys
- Removed V-Speed display below airspeed band
- Adjusted FLC and VS mode display in FMA
- Adjusted target speed display
- Added 1/2 Bank mode arc to PFD


mattnischan (approved):
@flymypi Sorry it took us forever to get to this! Definitely appreciate the contribution.

This looks good to me!

flymypi:
@mattnischan Thanks Matt!

I have one more item that is almost done, then I'll stop pushing to this branch.

flymypi:
**All the intended features for this branch are now completed.**

One FYI regarding the 1/2 Bank Mode Arc:
Display works as intended based on SimVar. However, the aircraft seems to spawn in 1/2 Bank mode by default. This was confirmed using the 0.8.0 release.

Maybe someone with a deeper understanding can take a look at that and maybe even implement the automatic mode switches the real aircraft does based on altitude?
@K20017 has references

mattnischan:
@flymypi
We noticed that the half-bank tooltip seemed reversed, but now I'm concerned that maybe it isn't. Can you confirm that the plane only banks 15 degrees by default on flight load? I'm not seeing that and I'm wondering if maybe that is totally ignored by the AP at the moment.

flymypi:
@mattnischan
It's not ignored because both default CJ4 and 0.8.0 WTCJ4 react to a mode change, just the wrong way round. According to the SimVar, they spawn at 15deg limit but bank to 30deg, then only bank to 15deg after the half bank  is deactivated.

I personally think it's a CJ4 specific issue, because in the default 747 bank limiting works fine (and with multiple increments).

dga711:
I am unsure. Can we merge this now? @flymypi

flymypi:
@dga711
I've got two more ideas I could take a look at later today. If there's no rush you can merge tonight, otherwise I'll just open a new PR for future stuff.

dga711:
@flymypi i opt to merge it now, so we can get it to QA before we release later today. thank you!

src/workingtitle-vcockpits-instruments-airliners/html_ui/Pages/VCockpit/Instruments/Airliners/Shared/WT/AttitudeIndicator.js
```diff
@@ -1594,6 +1607,9 @@ class Jet_PFD_AttitudeIndicator extends HTMLElement {
                     this.radioDecisionHeight.textContent = fastToFixed(val, 0);
                 }
                 break;
+            case "half_bank-active":
+                this.cj4_HalfBankActive = parseFloat(newValue) > 15;
```
flymypi, line 1611: Reverse this if you want to invert the arc display with the current state of the systems

dga711: i already noticed. will do thanks!

PR #621, opened 2020-12-17
Change wind to correct bus

### LINKED ISSUE #620: Wind Noise in Cockpit View
hi. 

i have muted some sounds and other sounds i have on 10/100. voices on 80/100. 
but in cockpit the wind noises are incredible loud and no way to make it silent. 

problem is since 0.8.0

K20017:
Puts wind back on the winds_generic_inside bus so it won't continue playing at game pause.

closes #620


dga711:
closes #620

mattnischan (approved):

PR #632, opened 2020-12-17
CJ4: Add visual runway extension fixes. Fix 3 letter airport entry. Fix LNAV fishtailing.

mattnischan:


mattnischan:
Also added a fix for a missing course to radial intercept leg type in LegsProcedure.

dga711 (approved):

PR #642, opened 2020-12-18
CJ4: 0.8.1 Release

mattnischan:
- [x] Update CHANGES
- [x] Update version in FMC ABOUT and package definition


dga711:
Former QA changelog
```
- Many PFD/MFD cosmetic changes contributed by TheFlieger (see https://github.com/Working-Title-MSFS-Mods/fspackages/pull/586)
- Fixed the option switch for FP Sync
- Fix FMS crashes when going to ARR DATA page
- HDG select line should not be showing permanently when in LNAV
- FMA mach value should always show 2 digits
- Removed unnecessary package that breaks G1000/(A32NX) compatibility
- Fix altimeter rounding error in SAI
- Fix fuel flow and weight in FMS Text on MFD
- Fix setting VFR runway in DEP/ARR page and flight plan handling
- Fix 3 letter airport entry
- Fix LNAV fishtailing when setting a Direct To while in a turn
- Fixed missing leg type mapping for a course to radial intercept type
- Click on origin/dest in FPLN should now copy the airport into the scratchpad
- Wind sound is now on the correct audio bus and should obey volume settings
- Input like "FL400" is now correctly handled as altitude restriction, not speed
```

cwburnett (approved):

PR #658, opened 2020-12-20
CJ4: Fix SimBrief handling of three letter airports

dga711:
SimBrief sends for ex. 3U2 as K3U2 and the sim doesnt like it.

If there is a number after K, it will strip the K and add a space at the end so the sim takes it.

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_FplnRecallPage.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_FplnRecallPage.js
@@ -4,7 +4,7 @@ class CJ4_FMC_FplnRecallPage {
         let json = "";
 
         let parseAirport = (icao) => {
-            if (icao.startsWith("K") && (/\d/.test(icao))) {
+            if ((/K.*\d.*/.test(icao))) {
                 icao = icao.substring(1).padEnd(4, " ");
             }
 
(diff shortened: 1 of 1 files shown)


src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC_FplnRecallPage.js
```diff
@@ -3,6 +3,14 @@ class CJ4_FMC_FplnRecallPage {
         let url = "https://www.simbrief.com/api/xml.fetcher.php?userid=" + pilotId + "&json=1";
         let json = "";
 
+        let parseAirport = (icao) => {
+            if (icao.startsWith("K") && (/\d/.test(icao))) {
```
Jessidhia, line 7:
```suggestion
            if (/^K\d/.test(icao)) {
```

mattnischan: I'm also confused on that. So if it's `KH2L`, we want to strip out the K also?

dga711: > I'm also confused on that. So if it's `KH2L`, we want to strip out the K also?

yes. the game wants "H2L "  it doesnt accept KH2L

dga711: > Actually, I'm not sure what you meant by "If there is a number after K". If it starts with K and there's a number anywhere, or literally if there's a number "after K"?

Thanks for the suggestion of doing it smarter. I will gladly take it ;)

dga711: So. The number can be anywhere in the ident after the K. So the test as suggested wouldn't work.
Needs optional other characters in between

dga711: How about 
image

mattnischan: Was just going to suggest that, yep. Looks good.

cwburnett (approved):

Issue #681, opened 2020-12-23
G1000, G5000 don't show full flight plans on MFD


BostonJeremy77:
Hi all. Starting with today's update, towards approach, the magenta/white lines of the flight plan is cut off short of airport - it still follows the last few waypoints, but the plan isn't shown on MFD. The flight plan in FMS is still fine.

modabasi:
In my case (TBM 930) the initial flight plan is shown, but when I load an approach to the flight plan, the purple lines dissapear on the MFD. The autopilot still flies the approach, but I can not see the path flown on the MFD.

Pascal207:
Hello.
I've had the same issue with the longitude.
When i load an approach, the flight plan dissapear on the MFD, and the flight plan on the fms have "blank" points.
The plane still follow the flight plan.

cywang90:
This should be fixed with v0.3.4.

PR #685, opened 2020-12-23
g3000 v0.3.4

cywang90:
- fixes the navmap not drawing approaches. (#681)
- PFD AP display now has correct flashing behavior
- moved custom WaypointLoader and XMLLogic to WorkingTitle directories


kaosfere (changes requested):
Can we leave XMLLogic.js where it is for now?   Like I mentioned in discord, that's a tricky one to move because it's read directly from VCockpit.html.   It _should_ be relocated, but it's been that way for ages without an issue, and moving it will make it hard for me to put out a bugfix for the A32x conflict.

kaosfere (approved):

PR #735, opened 2020-12-31
CJ4: Refactor PFD ILS/LDEV/VDEV

cwburnett:
- [x] LOC ONLY (doesn't show GS)
- [x] ILS (shows both LAT/GS when ILS is tuned in PFD)
- [x] VNAV/VDEV for GP or PATH
- [x] LDEV for RNAV APPR or when PPOS mode is enabled on the PFD


mattnischan (commented):
This looks a lot better but I think we can just delete all the commented out stuff.

cwburnett:
> This looks a lot better but I think we can just delete all the commented out stuff.

done and done

mattnischan (approved):

PR #741, opened 2020-12-31
Added new Baro Settings popup to the G3X.

arron-h:
This is activated by clicking on the baro reference value under the altimeter.
In order to bring this popup up, an ID was required to be added to the baro reference text in the shared Altimeter.js.

kingair_baro_set


arron-h:
Validation on the input is performed. As can be seen, the Enter button is greyed out. The Enter button activates when a valid pressure setting (ranging from 27.99 to 32.01) is input. Clicking Set Standard fills in 29.92 and activates the Enter button.

arron-h:
Other noteworthy additions: I've added a `title-small `CSS class to `gradientButton` as the default titles I felt were way too big.

Here's an example screenshot of the Transponder input page from the G3X manual, which I based the button from:

transponder

I hope to refactor other popup dialogs to use the same class for consistency.

arron-h:
And here's a single frame of a YouTube video of the Baro Setting page on a G3X Touch. Note that because the aircraft is on the ground, 'Set For Field' is displayed. Once airbourne I believe this changes to 'Set Standard' as included in this PR.

baro setting

arron-h:
Marking as draft as I'm doing a bit of rework to improve the clickspot handling, text alignment and code consistency.

arron-h:
OK, happy with this PR now. Attaching updates screenshot showing fixed centering of text.

kingair_baro_set_2

PR #778, opened 2021-01-06
Remove rate limiting of controller stepping

neoeinstein:
From my understanding, the use of rate limiting here was not intended. Including this rate limiting can have unintended consequences inside the controller in the event an update is skipped because 50 milliseconds hasn't yet elapsed. In such a case, the `deltaTime` passed to the controller may be significantly shorter than the elapsed sim time since the last update.

Options were to either accumulate the sim time between runs of the controller, or remove the rate limiting all together. This PR opts for the latter and removes the rate limiting.


dga711:
Thank you for finding this and informing us. It was a leftover from some previous testing.
We already changed it on the prerelease branch for 0.9.0.

PR #779, opened 2021-01-07
Fix fuel timing calculation and display

### LINKED ISSUE #568: Remaining fuel display shows 1:60 instead of 2:00 when recalculating
In the perf -> fuel mgt display. Not much more to say, really, when the remaining fuel hovers around an even hour, the minute indication goes to 60 rather than moving the hour indication up/down and the minutes going to zero.

neoeinstein:
Fixes #568


mattnischan (approved):

mattnischan:
This looks good to me. Thanks for the contribution!

PR #790, opened 2021-01-09
Correct hold fix calculation for magnetic variation

neoeinstein:
When the hold course is not true, then correct for the magnetic variation at the initial hold fix.

We could calculate magnetic variation at each fix, but it is unlikely that the variation will be extreme between hold fixes.

Images showing the 114M/R published hold at SEETS on the JFUND2 arrival into KBOS properly aligned with the arrival.
bildo
bildo


mattnischan (approved):
Looks good to me! Thanks for the contribution!

PR #807, opened 2021-01-11
Departure TA + PERF INIT > VNAV CLB to redirect to PG 1

Syntax505:
Noticed that you can't set the departure TA in ACT VNAV CLB but you can set the arrival TA in ACT VNAV DESCENT. Asked in the Discord and apparently it wasn't implemented so I decided to add it. This PR also sets the VNAV CLB option in PERF INIT to redirect to VNAV SETUP page 1 instead of page 3. (not sure if this is the case in real life but I'd assume that it is the case)


magiber4:
Hello, this change is good.
I can change TA in climb
but I think it is not effective in the sim
when I write new TA. (example 6000, in Spain it is when it changes)
the plane when it goes from 6,000 does not change TA to STD I have to do it manually
that's why I think that only the change is only visual in the fmc
is it possible to make it effective in the simulator?
thanks

Syntax505:
I believe it doesn't do it automatically in the CJ4, you'll have to manually set it to STD but I'm not 100% sure

cwburnett:
@Syntax505 Thanks so much for submitting this PR, we actually have addressed the VNAV CLB page in the next update v0.9.5 and will re-order the pages to go to CLB again before release. We really appreciate your participation and contribution, so while we won't merge this in since we already added storable values for the entire page, please don't let that discourage you from future contributions! Thanks!

PR #822, opened 2021-01-14
CJ4: Msg system and TypeScript page base

dga711:
- use new rollup build system for typescript
- changed external modules to declarations
- gets rid of manual exporting in wtsdk.ts
- page navigation service in ts
- page baseclasses in ts
- msg page implementation in ts
- new custom html element for pfd annunciations
- architecture for messages
- basic fmc,pfd annunciations implemented
- put previous pfd annunciations into the new structure (TOD, TERM etc)

known issues (to be resolve later):
- missing import of lzutf8
- CHECK ALT SEL is missing conditions, need to talk to @cwburnett 
- msg page "OLD MESSAGES" functionality missing


mattnischan (approved):

src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
```diff
@@ -845,6 +850,13 @@ class CJ4_FMC extends FMCMainDisplay {
                 break;
         }
     }
+
+    updateMsgs() {
```
mattnischan, line 854: Can we make this delta time based instead of a frame counter?

src/wtsdk/src/cj4/ui/fmc/CJ4_FMC_Page.ts
```diff
@@ -0,0 +1,42 @@
+export abstract class CJ4_FMC_Page {
```
mattnischan, line 1: I wonder if we could make this an interface instead and leave the internals and implementation to the specific pages.

mattnischan (commented):
So in general I dig this. I think one issue I see as we add more messages is that the message system right now is all pull based. So each message has no real knowledge of its own context and just has to poll forever.

I think ideally only the FMC would be in charge of messages. Each system (LNAV, VNAV, etc) could then call something like:

```typescript
//from inside HoldsDirector, lets us capture 'this' and other locals
this.fmc.postMessage('EFC TIME ELAPSED', context => {
  return new Date() > this.getEfcTime(holdIndex);
});
```

Then, in the FMC message controller, it would do something like:

```typescript
public postMessage(text: string, handler: any => boolean) {
  this.activeMessages[text] = {text: message, handler, context: {}};
}

public update() {
  const messagesToRemove = [];
  for (var messsage of this.activeMessages) {
    if (!message.handler(message.context)) {
      messagesToRemove.push(message);
    }
  }

  for (var message of messagesToRemove) {
    delete this.activeMessages[message.text]
  }
}
```

src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
dga711: yeah we should replace all those in cj4_fmc with time based values

src/wtsdk/src/cj4/ui/fmc/CJ4_FMC_Page.ts
dga711: while interface is always good too ofc. i want to provide the pages with the 90% case base implementation... ?

dga711:
> So in general I dig this. I think one issue I see as we add more messages is that the message system right now is all pull based. So each message has no real knowledge of its own context and just has to poll forever.

This would be a totally different approach then. Which is valid too of course.

My thinking was that I want the message system to be independent. It is checking the "status" of the aircraft and posts message when conditions are met.
I didn't want the system implementations to have to care about when to send which message and also clear it.
Guess we can discuss this further in voice when you have time.

src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
dga711: as we should resolve this in general, this should go to a different pr

PR #845, opened 2021-01-21
CJ4: Ghost LOC/GS on PFD

cwburnett:
CJ4: Ghost LOC/GS on PFD


mattnischan (approved):
This seems to test OK, looks good to me.

PR #921, opened 2021-02-09
CJ4: Datalink

Syntax505:
This PR aims to (partially at the moment) simulate the Collins Corporate Datalink System CMU-4000 which is already present onboard the Working Title CJ4 (albeit disabled). Its current features include the ability to request METARs/TAFs/ATIS, an option for realistic processing time (15 seconds), and options to select different sources for METARs and ATISes. In the future another PR will be made to simulate the DEPART CLX and OCEANIC CLX options as well.

Note: An acars.json file is currently present in the html_ui > Pages > VCockpit > Instruments > Shared > Utils directory. This is for the future implementation of departure clearances and oceanic clearances via Hoppie as written above.


dga711:
Thanks for the PR. Looks good in general.

But I don't think we should just go in and use FBWs APIs. Not really a nice thing to do.
And although they could be open to us using it, I'd rather like to see it being independent of any changes they may do in the future.

Syntax505:
Then I shall look for better solutions.

Benjozork:
All good on using the API - WXR is a minimal load on our servers.

mattnischan:
Merging this into our 0.10.3 prerelease branch. Much thanks for the contribution!

PR #933, opened 2021-02-18
Added rounding to engageFlightLevelChange speed

collinkoldoff-KF:
This is my small contribution to WT and this is the only thing I could figure out how to fix.


K20017:
What exactly is this fixing?  From our knowledge, everything with FLC appears to be working.

PR #950, opened 2021-02-24
Fix for CRS not auto-tuning bug

Syntax505:
Currently when an ILS approach is selected in the FMS, pressing APPR and having the nav source automatically transferred does not automatically change the CRS of the needles. This fix rectifies that by checking that the frequency was indeed properly transferred instead of checking for if it has failed.


Syntax505:
Closing this as @K20017 seems to have fixed it in his latest PR.

PR #968, opened 2021-03-05
CJ4-tunepg-updates

slip664:
-Minor cosmetic changes to match real aircraft
-Modified PAGE 2/2 to remove ADF 1 and ADF 2 to match real unit and prevent user confusion
-Added ATC CONTROL page
-Added Dispatch page when avionics switch set to DISPATCH mode. - WIP

-Added code for ATC CONTROL and DISPATCH mode - needs review.


slip664:
Thanks, Matt. Should be good to go now.

dga711 (approved):

PR #975, opened 2021-03-08
CJ4 Pre 0.11.0

dga711:
Changelog:

1. Navigraph MFD Charts implementation & fixes
2. PFD Baro preset function added.  (After STD is pressed on Baro knob,, a preset value can be selected and swapped with STD press again once passed transition altitude)
3. Added VNAV window to FMS text (Must be selected on DSPL MENU vnav window)
4. Added Advisory Descent (DES instead of TOD) when no arrival or approach is loaded, will give you an advisory point to start descent to reach 1500' AFE 10nm from airport.
5. Added TUNE page functionality - ATC Control page, dispatch function, format changes, 
6. Fixed vnav bug where below constraints were not observed in the first segment of the vertical path.
7. Fixed glideslope capture so it will only go active when within 1/4 dot or so (Used to go active at full deflection)
8. Fixed another vnav bug where below constraints on SIDs further than 40nm were not seen.


K20017:
Changelog:

1. Addition of Navigraph charts implementation

cwburnett (approved):

PR #1003, opened 2021-03-12
Added custom cockpit and cabin textures for the CJ4

MoreRightRudder:
Co-Authored-By: Tyler Knox <16512508+tyler58546@users.noreply.github.com>
FlightSimulator_2021-03-11_20-59-51
FlightSimulator_2021-03-09_15-51-15
FlightSimulator_2021-03-08_03-17-43
FlightSimulator_2021-03-05_15-43-32

There are new cockpit models as we had to merge the new glass object into the cockpit. The FMS also has a bezel positioned so as not to interrupt any FMS Text. 

There are new cockpit and cabin textures. Cockpit colors are matched to the Gen2 version of the aircraft and the cabin is the Jet Black Premier Collection offered by Cessna.


dga711:
Great work. Looks really cool ingame.

Saw a few issues with reflections or something. Like this 
https://streamable.com/0fc2t5


Two other topics
- We've been quite happy with not having to override the cockpit model anymore. Because it was very prone to break after updates
- The filesize of all of this huge (~900MB) compared to our current package size of around 50MB
So we were thinking about offering it as an additional package on GitHub and Discord instead of having it default with our mod download. So people don't have to download 900 megs it with every small patch we offer for the CJ4. How do you feel about that?

dga711:
On some zoom levels the glass starts flickering
https://streamable.com/ffkmfa

MoreRightRudder:
> On some zoom levels the glass starts flickering
> https://streamable.com/ffkmfa

Thank you for checking it out dga! I will take a look at these issues and see if I can resolve them. The carbon fiber wrap is a tough look to master as is it wrapped in a plastic coating in the real plane. The sim interprets this as reflective and applies screen space reflections to it which have an odd effect at certain viewing angles. I completely understand why this would not become part of the current package at this stage and would be happy if it were offered as an additional package! Thank you so much!

cwburnett:
From looking at the files only and seeing no overlap at all, it looks like this could easily sit as an additional optional mod right next to the cj4. That seems like it would be a really great option. Very cool 👍

cwburnett:
@MoreRightRudder - so I packaged up the textures and model as a separate package for the community folder and tested; all went well with that, so I think offering this as an additional package will work really nicely. And by the way, I really liked the enhancements. 👍  The only issue I had was that the additional texturing on the PFD/MFD made daytime visibility a bit of a challenge, so I think we either need to find a way to make the PFD/MFD brighter during the day (increase the max brightness) or reduce some of the texturing on the PFD/MFD. For whatever reason, when looking in full daytime, the FMC is still very readable, but when looking closely, there's definitely more lcd texturing on the PFD/MFD than the FMC, so the FMC was crisp and easy to read, but the PFD/MFD were harder to read when zoomed out. I'm not sure what the best solution there is, but in any case, I think whenever you're ready, we can make the texture mod available along with the core package in releases, unless you want to make it available in a different way.

windhover-wi:
I'm loving the textures, especially the front dash and throttles (well, really the whole throttle console). My only niggle is something Chris has already mentioned above with the PFD/MFD and standby display all looking overly LCD-ish and getting really washed out in daytime. Other than that I will definitely rock this moving forward.

MoreRightRudder:
> @MoreRightRudder - so I packaged up the textures and model as a separate package for the community folder and tested; all went well with that, so I think offering this as an additional package will work really nicely. And by the way, I really liked the enhancements. 👍 The only issue I had was that the additional texturing on the PFD/MFD made daytime visibility a bit of a challenge, so I think we either need to find a way to make the PFD/MFD brighter during the day (increase the max brightness) or reduce some of the texturing on the PFD/MFD. For whatever reason, when looking in full daytime, the FMC is still very readable, but when looking closely, there's definitely more lcd texturing on the PFD/MFD than the FMC, so the FMC was crisp and easy to read, but the PFD/MFD were harder to read when zoomed out. I'm not sure what the best solution there is, but in any case, I think whenever you're ready, we can make the texture mod available along with the core package in releases, unless you want to make it available in a different way.

Thank you! I am glad you guys like it! The glass on the PFD and ND don't have extra texturing per se, rather a different tint. It is the purple tint which darkens the display that I can change. The tint is currently at 1% strength, so my only option is to change the color of that tint to a lighter shade of purple which I can easily do. In the meantime, I am down to allow you guys to package this into a link on discord or however you wish to distribute it as it stands now. Currently I am flying on a trip and will be back late tonight. Changes to the glass could be made within 2 days time. Thank you for your time and feedback!

PR #1048, opened 2021-03-24
CJ4 Pre-Release 0.11.4

### LINKED ISSUE #1064: [CJ4] VNAV not meeting "At or below" constraints
EAGUL6 arrival into KPHX, starting at GUP. Cruise altitude FL400.
msedge_nlIfaloghc
First constraint is a FL330B at TINIZ. This screenshot shows ALT selector 7000, VNAV armed. At this point, no runway is selected, haven't picked up ATIS, and ATC hasn't advised whether landing east or west:
explorer_sByARKzyWZ
Now expecting "landing east", so RWY 7R selected, and the FMS loads the appropriate fork of the STAR. Note how the TOD has moved (this seems to be where the problems begin):
explorer_5qUIcuB4b3
Approaching TINIZ, VNAV has just started the descent, so not going to make the FL330B constraint:
explorer_geM8uk0SzW
PAYSO has a FL240A/FL300B constraint, crossed at just below FL330:
explorer_h326FD3RKE
The next point with an "At or below" constraint on the arrival is HOMRR 17000B, here about to cross HOMRR, well above 17000:
explorer_IVRIXq5X8N

At this point I stopped the test. This was a recreation of a prior flight so I could gather screenshots, so the result was the same on two different flights.

cwburnett:
- [x] Add ILS APPR Mode without approach loaded into the FMC
- [x] Fix final approach fix calculation in LNAV to discard MAP waypoints <3nm from runway
- [x] Fix VNAV stopping climb when no SID is selected but a STAR is
- [x] Fix VNAV Text display to show descent data in SID after passing all climb restrictions
- [x] Departure Runway undefined fixes (incomplete @dga711 @K20017 ??)
- [x] slip's terrain changes and 180 degree mask
- [x] Panel Light Knob
- [x] ROSE/ARC Fixes (@K20017 more to do)
- [x] Adjust min wind vector/speed
- [x] Fix idle thrust & 75% thrust error
- [x] Arm LNV on ground, will engage at 400'
- [x] Vertical Direct will allow second arm
- [x] Vertical Direct to TO fix
- [x] Adjust display of active DTO wpt on direct page
- [x] Adjust FD behavior to disable all AP modes when AP is off and FD is turned off (this required a new HVAR - needs to be added to the HVAR list)
- [x] Added pitch ref value to PTCH FMA display
- [x] Fix DES waypoint appearing even after it has been passed
- [x] Adjust VNAV button behavior to only disable VNAV button when in GP/GS, instead of when in APPR lateral mode.
- [x] Fix bug with VNAV calculations after a DIR TO (closes #1064 )
- [x] Adjust gps track bug on MFD/PFD and enable during ground ops
- [x] MFD/PFD add rose cardinal direction and 45 degree arrows
- [x] Fix bug with missing some A/B or A/ constraints in the first segment in some rare cases
- [x] Added feature to break the path at the first constraint so the descent path to the very first constraint is the desired FPA set in VNAV DESCENT SETUP PAGE
- [x] Fixed bug with Vapp/Vref alignment/showing '0' value instead of blank
- [x] Adjusted AP behavior when PATH changes while PATH is active; if an out of PATH situation occurs, AP mode changes to PTCH.
- [x] Implement runway slope of -2% to +2% for takeoff and approach ref pages
- [x] Fixed FMS bug where departure runway altitude appears on LEGS page and VNAV window
- [x] Implement landing factor distance on Approach Ref page
- [x] Fixed bug where selected altitude would go below 0' and not display - now the sim AP value will not go below 0'.
- [x] Added ability to enter custom/new pax weight on PERF INIT page using format "/XXX" or "5/XXX" and the pax weight value is persistent.
- [x] Added ability to enter Gross Weight on PERF INIT page to override ZFW and/or Pax/Cargo inputs.
- [x] - implement custom vertical autopilot
- [x] - added altitude capture for PTCH
- [x] - added altitude capture for TO/GA
- [x] - streamlined vnav code because we no longer have to manage sim alt slots
- [x] - improved custom vertical AP: adjusted alt intercept logic
- [x] - fixed VNAV window when only a climb is loaded
- [x] - fixed units display problem with baro present window
- [x] - fixed VNAV bug with calculating path in certain circumstances
- [x] - disabled FD mode reset on landing
- [x] - fixed range banana only reads from selected altitude
- [x] - note: the ALT button is now tied to H:WT_CJ4_AP_ALT_PRESSED and the emissive lvar is L:WT_CJ4_ALT_HOLD

TODO:

- [x] Change log
- [x] Update version numbers
- [x] Update website (@flymypi )
- [x] Update simvar list for WT_CJ4_AP_FD_TOGGLE


flymypi:
Once we use the new website, changelogs can be put here:
[Website Github CJ4 Releases](https://github.com/Working-Title-MSFS-Mods/working-title-msfs-mods.github.io/tree/master/docs/cj4/_posts)

Only adaption that will need to be made is that the date needs to be included in the changelog name.

Version numbers and latest releases etc. then update automically on the website.

dga711 (approved):

PR #1057, opened 2021-03-28
[G3000] Fix "NO WIND DATA" font size on PFD

Nezz:
The font size was too big, so "DATA" actually ended up being printed on the top of the screen. Once the font size is adjusted the text is displayed in two rows, which is how it is on the Garmins.

Before:
Before
After:
After


cywang90:
I actually fixed this by changing the font type to what it was originally intended to be (forgot to update the css file before release).

PR #1149, opened 2021-04-28
fix(cj4): only use lbs paxWeight for zero fuel weight

### LINKED ISSUE #1102: CJ4 - PERF INIT ZFW Calculation
Hi there.

I’ve encountered a weird problem in 0.12.0.  When entering the PAX and Cargo details on the PERF INIT page the ZFW calculation doesn’t seem to calculate properly.  Prior to 0.12.0 I could match up the ZFW in the FMS with the MSFS loading screen by using the same PAX and cargo info but now it doesn’t align.  To see an example of what I mean if you try adding 1 PAX @ 77kg into the PERF INIT page of the FMS, the calculated ZFW only increases by 35kg.

I get round this by just entering the ZFW directly but does seem like it might be a hug?

cheers
Donald

thor:
paxWeight was used both for calculating the zero fuel weight, but also for displaying the weight per passenger.
While the weight per passenger would display correctly, it would be calculated incorrectly with a lower weight due to all calculations being done in lbs.

This commit postpones conversion of paxWeight until displaying the value, which seems to be the norm.

Fixes #1102. Hopefully.


K20017:
👍

K20017 (approved):

PR #1159, opened 2021-05-03
Updated CJ4 lights

Uwajimaya:
First off, congrats on officially joining the MSFS team, well deserved! Ok, here's what's new: 

- Removed all lens flares, they just don't look good enough with how the current lighting is implemented. 
- Created a new tail beacon (made up of 3 parts) that's more realistic than the standard "omni" light. 
- Re-timed the wing strobe and tail beacon. They alternate evenly during the cinematic, and when turning all lights on simultaneously. If you can figure out way for that to occur when using cockpit light switches, that'd be awesome. 
- I did NOT combine the white nav lights into one, they remain left/right. You guys did that, so if there was a purpose... keep it. If not, it can be combined into one file. 
- Finally, I made the logo light "cone" smaller - before that, it was illuminating the tail and the ground underneath it in a wide area. Now it's illuminating more of the tail and not much else.

Note - the "texture" folder inside the effects folder can be removed entirely, I didn't know how to do that in a PR.


dga711:
Hey there. Thank you for the wishes and the contribution.
Will check it out in the coming days and pull for an upcoming release

PR #1173, opened 2021-05-13
[CJ4] Allow NAV2 course to be set when NAV1 NAV_TO_NAV_TRANSFER_STATE is Armed

icalder:
Thanks to everyone involved in this project.  I'd pretty much given up on MSFS for now and gone back to X-Plane, but the CJ4 mod is just so immersive that it's my goto now.

I found that I could not change NAV2 course with the FMS set up for a flight plan with ILS arrival.  Some of my routes have DME arc arrivals and I like to set up the FMS with auto NAV1/ILS, then use NAV2 for the arc procedure in heading mode before switching back to FMS  for final approach.  Sometimes also it is useful to configure NAV2 for a missed approach course.

I found that when WT_NAV_TO_NAV_TRANSFER_STATE changes from 0/1 to 3(Armed) it is no longer possible to change NAV2 course.  This patch allows NAV2 to be changed when NAV1 transfer is Armed.  I'm not sure if that is allowed in the real aircraft but I can't see a reason why it wouldn't be.


dga711:
Sorry, this was sitting here for so long.
Someone will take a look and we merge it for the next release. Thank you very much!

dga711:
@K20017 @cwburnett can you review this please?

PR #1217, opened 2021-06-01
CJ4 Altimeters Touchup

charles-vomacka:
Details in attached PDF.

[Altimeters_Touchup.pdf](https://github.com/Working-Title-MSFS-Mods/fspackages/files/6580090/Altimeters_Touchup.pdf)

### DIFF
--- a/src/workingtitle-vcockpits-instruments-airliners/html_ui/Pages/VCockpit/Instruments/Airliners/Shared/WT/AltimeterIndicator.js
+++ b/src/workingtitle-vcockpits-instruments-airliners/html_ui/Pages/VCockpit/Instruments/Airliners/Shared/WT/AltimeterIndicator.js
@@ -99,7 +99,7 @@ class Jet_PFD_AltimeterIndicator extends HTMLElement {
         this.cursorIntegrals.push(new Avionics.AltitudeScroller(3, 52, 1, 10, 1000));
         this.cursorIntegrals.push(new Avionics.AltitudeScroller(3, 52, 1, 10, 100));
         this.cursorIntegrals.push(new Avionics.AltitudeScroller(3, 52, 1, 10, 10));
-        this.cursorDecimals = new Avionics.AltitudeScroller(5, 25, 10, 100);
+        this.cursorDecimals = new Avionics.AltitudeScroller(5, 25, 20, 100);
         if (!this.rootGroup) {
             this.rootGroup = document.createElementNS(Avionics.SVG.NS, "g");
             this.rootGroup.setAttribute("id", "Altimeter");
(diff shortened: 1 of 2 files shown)


slip664:
This is a good idea. I should do something like this for the SAI. Very nice!

charles-vomacka:
@slip Thanks. I have not seen a reference for the SAI that shows this, but I would expect that it is implemented this way to be consistent.

slip664:
The graphic is slightly more complicated but it's a similar effect.

saiground

charles-vomacka:
@slip664 Yea, you may have to implement it using a path instead of a simple rect. Would look cool though.

charles-vomacka:
@slip664 Actually, don't know if you could do that with a path.

charles-vomacka:
@slip664 If you are going to implement as I have, just borrow the code.

flymypi:
Do you know if the blocks scroll on the real thing?

We have a solution prepared using the native scrollers but it involves changing our font.

If they don't scroll we don't need to change the font anymore, if they do, this is still a very nice temporary fix.

charles-vomacka:
@flymypi No, none of the references I saw showed it changing. All of the references that I used were from the same Youtube account -  Mehr Spaß am Spaß Haben Haben. I will check a couple others - Citation80HBflyer and Premier 1 Driver. I know that Premier 1 Driver uses a different PL21, but a lot of the functionality is the same - it may provide a hint of the correct behavior.

charles-vomacka:
@flymypi  In my opinion, the font is more important that the scrolling. The font is distinctive with its open 4 and I think with the correct font it would look more true to the real unit.

charles-vomacka:
@flymypi I found a reference - https://youtu.be/LSdsIvtex0M?t=2162
It shows it scrolling into position.

charles-vomacka:
@flymypi I will research SVG animation and see if I can implement the scrolling effect, then the font would not have to be changed. Would this be acceptable, or are the native scrollers more efficient?

flymypi:
@charles-vomacka I agree on the font fixes and the native scrollers would be a much better solution because it is triggered either way. 

That being said, font editing requires costly software which is why there is no one working on it right now.

charles-vomacka:
@flymypi Oh, when you said change the font, I thought you meant to a different font. I will explore FontForge, It is an open source font editor - although I have read that it has a steep learning curve. Is editing the Collins ProLine font OK? If so I will try to add a box character to the font.

mattnischan:
@charles-vomacka I originally started to use FontForge to create the Collins ProLine font, but ran into a number of problems that made it really hard to do the font in it (just really didn't have the relative measurement tools required for all those curves, not to mention it loves to crash frequently). So, I ended up doing it in FontLab, which is unfortunately not free and I haven't renewed the license since. The FontLab source is in the repository, though, and it should stay in sync with the actual typeface.

charles-vomacka:
@mattnischan OK. I was rushing out the door earlier, so did not have much time for research - not sure if FontForge allows importing and editing a font. On my drive, I thought about the problem and was going to use the Collins Proline font as the basis for custom font that replaced the zero with a box - if this is possible. The new font could be used for the ten-thousands and thousands columns of the altimeter; no other changes required. So, I should not do this?

charles-vomacka:
@flymypi @mattnischan I will try to change my implementation using a customized font. Then this can be used as a temporary fix until the team has time to implement a better fix

PR #1221, opened 2021-06-05
CJ4 Altimeters Touchup (scroller)

charles-vomacka:
Second version of Altimeters Touchup. In this version, the leading zeros in the ten-thousands and thousands columns are replaced by a block that scrolls into position. You guys can decide which version, if either, you want to pull.

I tried to implement this so that, hopefully, it will be easy for the team to implement a customized version of your font. I created a whole new font (from your font) with a customized $ character in the shape of a block. This will allow you to just get rid of my custom font when ready, and my font can't be confused with your font. If you modify the $ character in your font, then you should only have to remove two font entries in PFD.css and everything should work. If you customize a different character, then the update call to CJ4_AltitudeScroller will also have to be changed. Additionally, if a custom font is created for the SAI, then it should be able to implement this also.

Issues:
1. I tried to override the update method of the AltitudeScroller class, but could not get it to work right. So, I recreated the entire class and then modified it.

2. I could not find the layout.json file anywhere in source. An entry for the custom font will need to be added to it.


src/workingtitle-vcockpits-instruments-airliners/html_ui/Pages/VCockpit/Instruments/Airliners/Shared/WT/AltimeterIndicator.js
```diff
@@ -1760,9 +1760,15 @@ class Jet_PFD_AltimeterIndicator extends HTMLElement {
             //         this.cursorIntegrals[1].update(_altitude, 1000, (hideZeros) ? 1000 : undefined);
             //     }
             // }else{
```
flymypi, line 1762: Have you tried uncommenting this code and check whether it scrolls?

The `clear`-method is built in and I've had it working with that block of code back in January. The only thing I'm not sure about is whether it scrolled or just appeared.

If it does scroll, there is no need to manipulate the `CJ4_AltitudeScroller` class, in my opinion

charles-vomacka:
@flymypi If I uncomment the code, it replaces the 0 with an X. The X does not scroll in, it just appears.

flymypi (approved):
In that case you can remove those legacy comments and I think we're good.

charles-vomacka:
OK, great.  Thanks.

Did you see my comment about needing to update layout.json?

The only change I made to the class was to add a parameter to the update method and then use that parameter.

flymypi:
Afaik, the layout.json is automatically generated with every build, so don't worry about that.

charles-vomacka:
@flymypi 
OK, thanks.

PR #1247, opened 2021-06-29
CJ4: Pre 0.12.6

dga711:
- [ ] QA
- [x] Update Versions
- [ ] Release Notes (see milestone)

TODO
- [ ] Do flights. Test AP modes and especially button interaction
- [x] FMA says "EMSP" when in Pitch (this may be because browser version changed? It is meant to be the em-dash length whitespace - we may need to find another way to add whitespace)
- [x] Some stuff is now missing animations like CCP Knobs push
- [x] Lower CCP Knob push doesnt seem to work
- [x] CCP Joystick animation borked
- [x] Engine Starters not working
- [x] Throttle animation borked (xml changed a lot)
- [x] Pitot Heat buttons work, but CAS still shows up. Do the HEAT and DEICE Buttons actually set the vars?
- [x] New AS3000 template makes pfd/mfd brightness not work anymore
- [ ] Reevaluate and check all autopilot related modelbehaviors (they are all blinking...need put new interaction tags in)
- [ ] Pressing B/C throws error in FMC
- [x] I think FD Toggle is broken
- [x] VNAV button does nothing
- [x] Audio Panel is screwed
- [ ] Some light buttons missing sounds?
- [ ] Lower CCP knob LONG push doesnt work anymore for reselecting charts
- [ ] Had to replace AS3000.xml, so we need to put our wwise events back in @K20017 
- [ ] MASTER/CAUTION buttons weirdly blinking when mouse over it
- [x] YD toggle
- [x] chart panning knob is broken (only in LEGACY, asobo has to fix)
- [x] setting baro std only sets the QNH to 1013 (i.e. doesn't show 1013 STD but rather 1013 HPA)
- [x] the standby flight display's qnh is also not synced with the pfd qnh
- [x] baro preselect doesn't work
- [x] Mach switchover from ias to mach goes to mach .37 ??
- [x] Engine Anti-Ice (SimVar.GetSimVarValue("ENG ANTI ICE:1/2", "number"); does not get set with WING/ENG ANTI-ICE buttons
- [x] Warning in console about NetBingMap radius


charles-vomacka:
@dga711 
Because of merge, line 1077 of CJ4_PFD.css needs to be changed to 
fill: var(--green)

dga711:
> @dga711
> Because of merge, line 1077 of CJ4_PFD.css needs to be changed to
> fill: var(--green)

Hm. I thought I cared about those during merge. Thanks for checking it out. Will fix 👍

cwburnett (approved):

PR #1299, opened 2021-08-03
temporary fix G3000 SVT Terrain/Compass

synqark:
This fixes about G3000 are:
- SVTTerrain (also Flight Path Vector) bugfix since v0.7.4
- Some Compass issue fix (I cant specify what PFD_Compass affects)

Perhaps this is temporary fix until use `CommonPFD_MFD.js` again.
I tried to use merged `CommonPFD_MFD.js` . but it was difficult to me :pray:


dga711:
@cywang90

PR #1318, opened 2021-08-08
Skip unrecognized waypoints when importing Flightplan from SimBrief

billtt:
Skip unrecognized waypoints and continue when importing Flightplan from SimBrief


billtt:
When I'm flying long-distance flights, I often have the problem of importing process (from SimBrief) being stopped because there were unrecognized waypoints. I made a small change in the code to skip those unrecognized waypoints and continue with the rest of the waypoints.

dga711:
Looks good to me. Thank you!

dga711 (approved):

PR #1345, opened 2021-08-24
Black Screen Fix

an3k:
Removed reference for FileHandler.js (since that file isn't shipped) and based on a chat in Discord "not used anymore".


cwburnett:
@an3k Thanks testing this out now!

an3k:
@cwburnett Can we get this merged please? Yes, there are only a very few people still using the non-NXi variant but still there are very few people using it. Since it's such a simple fix and would make the mod "ready for archival" I'd love to see it merged :)

PR #1404, opened 2021-10-21
Release/g3000 v0.7.6

cywang90:
* Fixed attitude indicator not updating.
* Fixed error when activating Direct-To with no flight plan.
* Fixed flight path marker not displaying.


dga711 (approved):

src/workingtitle-vcockpits-instruments-navsystems-g3000/html_ui/WTg3000/Avionics/Shared/PFD/Main/G3x5_PFD.js
```diff
@@ -554,10 +557,18 @@ class AS3000_PFD_MainElement extends NavSystemElement {
 }
 
 class AS3000_PFD_Attitude extends PFD_Attitude {
-    constructor(instrumentID) {
+    /**
+     * @param {String} instrumentID
```
Jessidhia, line 561: (if you're using typescript for type information) you should use lowercase `string` for strings, as `String` would be a boxed object created with `new String()`; same with `Boolean` (and `Symbol` if it had a constructor)

Nezz: These changes are coming from Asobo

PR #1409, opened 2021-10-24
Fix annunciator buttons on the Longitude

Nezz:
Now that the annunciators are working on the Longitude with SU6, it became really bothersome that the buttons did not work. After some digging I found that the issue is that the logic for managing the annunciations is in the `Cabin_Annunciations` class. The caution/warning reset buttons work on the TBM because its MFD uses that class (in the form of `Engine_Annunciations`). The Longitude however does not have annunciations on the MFD. It needs to use `Cabin_Annunciations` on the PFD instead.

The end result:
image

Best of all, you can finally mute the warning chimes.


Nezz:
@cywang90 I believe my PR got closed because https://github.com/Working-Title-MSFS-Mods/fspackages/pull/1404 was merged. Can you re-open it by changing the target branch to `main`?

PR #1411, opened 2021-10-25
G3000 annunciator fixes

Nezz:
**Longitude changes**
Fix "Master warning reset" and "Master caution reset" buttons. Now that the annunciators are working on the Longitude with SU6, it became really bothersome that the buttons did not work. After some digging I found that the issue is that the logic for managing the annunciations is in the `Cabin_Annunciations` class. The caution/warning reset buttons work on the TBM because its MFD uses that class (in the form of `Engine_Annunciations`). The Longitude however does not have annunciations on the MFD. It needs to use `Cabin_Annunciations` on the PFD instead.

The G5000 PFD also did not support blinking annunciations. That's supported now by adding the necessary entries to the CSS.

End result:
image

Known issues: The aircraft model has switched up the two the reset buttons. According to a Textron employee:
>Amber Outboard, Red Inboard is correct, but the wording is wrong, Warning should always be closest to the Pilots. They just need to change the wording.

**TBM changes**
CAS messages should not be displayed on the PFD under normal display conditions (e.g. no reversionary mode). End result:
image


roadfox (commented):
Can only test the TBM 930 as this is the only aircraft i know well enough.
Anoing Gear down alarm which was not silent after acknowledge is now working correct.
CAS Messages are now only shown in MFD and not in PFD as i would expect.

grafik

thanks allot

PR #1417, opened 2021-11-05
CJ4 - Display brightness turned up to eleven

charles-vomacka:
Increased the maximum display brightness. Displays are now legible even in the brightest, most glary conditions. Users are still able to reduce brightness with the brightness control.

### DIFF
--- a/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
+++ b/src/workingtitle-vcockpits-instruments-cj4/html_ui/Pages/VCockpit/Instruments/Airliners/CJ4/FMC/CJ4_FMC.js
@@ -77,7 +77,7 @@ class CJ4_FMC extends FMCMainDisplay {
         SimVar.SetSimVarValue("L:WT_CJ4_INHIBIT_SEQUENCE", "number", 0);
         SimVar.SetSimVarValue("L:WT_CJ4_TFC_ALT_ABOVE_ENABLED", "number", 1);
         SimVar.SetSimVarValue("L:WT_CJ4_TFC_ALT_BELOW_ENABLED", "number", 1);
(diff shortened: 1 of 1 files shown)


dga711:
Will check how it looks in the sim. Although back then I thought that even a value of 1.0 already looked way too bright.

charles-vomacka:
@dga711 In my testing the displays, to me, look like they are the proper brightness in daylight conditions. At night, if the brightness control is at max, then the display is way too bright - washed out. The user has to adjust the brightness down. I tested in the brightest and darkest conditions I could find. This testing indicates that the range of the brightness control matches the range of low to bright conditions well. My thinking is that this change puts control in the users hands; they can adjust the brightness to meet conditions. Currently, it is impossible to adjust the display to be bright enough in bright conditions.

I will understand if this PR is not accepted - it is definitely a kludge with some unwanted side effects. The real fix is for the sim to not darken the displays. I have not been able to figure out what is being simulated by darkening the displays. This is the only sim I have ever used that purposely make the instruments hard to read. This makes no sense to me. A lot of time and money has been, and is still being spent on developing these displays, only to have the sim then make them unreadable.

charles-vomacka:
@dga711 An additional note. I did all of my testing in SU6. There may be changes in SU7 that would change the results.

charles-vomacka:
@dga711 An easy test, in daylight conditions, is to pop-out the PFD display (right ALT, left mouse click the display). With the display popped-out, the sim does not darken it and the brightness control has no effect on it. This makes it easy to compare the display as it was constructed (no darkening) to the in-panel display. To me, this change makes the in-panel display, while in bright ambient conditions, look like the popped out display. In other words, it looks like it was intended to by the creators.

dga711:
Will pull it in and then adjust as we go. Thank you

charles-vomacka:
Hopefully they will make some adjustments in sim to mitigate the issue some.

PR #1478, opened 2022-01-17
CJ4: Engine Improvement

charles-vomacka:
Adjusted Fuel Flow Controller PID to minimize overshoots/undershoots. The overshoots/undershoots are what cause the engines to surge during throttle changes. I am marking this as Draft while I continue testing and hoping that others will try it out and see if they encounter any issues.

### DIFF
--- a/src/workingtitle-aircraft-cj4/SimObjects/Airplanes/Asobo_CJ4/engines.cfg
+++ b/src/workingtitle-aircraft-cj4/SimObjects/Airplanes/Asobo_CJ4/engines.cfg
@@ -87,9 +87,9 @@ high_fuel_flow = 1400
 min_n2_for_fuel_flow = 12
 mach_influence_on_n1 = 10
 fuel_flow_max = 1600
-fuel_flow_controller_p = 0.123
-fuel_flow_controller_i = 0.007
-fuel_flow_controller_d = 0.145
+fuel_flow_controller_p = 0.15
+fuel_flow_controller_i = 0.008
+fuel_flow_controller_d = 0.08
 fuel_flow_controller_iboundary = 2
 fuel_flow_controller_dboundary = 100
 max_torque_protection = 0 ; 0 = no protection, value of the torque triggering a protection by automatically limiting the fuelflow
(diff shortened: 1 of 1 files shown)


charles-vomacka:
Would especially like K20017's feed back - is throttle responsiveness seem fairly realistic.

charles-vomacka:
I am taking this PR out of Draft.

There seems to be an issue with the engine configuration that does not allow the Fuel Flow Controller PID to be adjusted correctly. The graph below show fuel flow as the throttle is increased from idle to high cruise. The X axis is time in .2 second increments. The Y axis is Corrected Fuel Flow in lbs/hr. 
cruise

As you can see, the fuel flow increases to fuel_flow_max (1600), stays there for approximately 1.6 seconds (this varies slightly between test runs), and then comes down to the commanded value. This is because the engine is trying to pull the fuel flow well above fuel_flow_max.

I temporarily increased fuel_flow_max in the configuration file, and then ran some tests. The following graphs show the results (X axis is in .5 second increments in these graphs).






As you can see, the engine is trying to pull fuel flow to about 1800 – sometimes as high as 2000.

This behavior can be corrected with the Fuel Flow Controller PID by turning the gain (P) way down. This slows the Fuel Flow Controllers response to a level that does not allow the fuel flow to increase all the way to 1600 before the engine is done trying to pull it too high. Decreasing the gain in manner is not a viable solution. It causes the throttle to not be responsive enough – it takes forever to go from idle to takeoff power. Because of this, my PID values are a compromise between responsiveness and minimizing overshoots. With these new PID values it takes approximately 1.5 to 2 seconds longer than it should to go from idle to takeoff power, but (except for the overshoot describe above) reduces overshoots to approximately 1% N1 (you will really only see these N1 overshoots at %N1 values below 50%). You will have to decide if this is an acceptable compromise. I have also noticed that on rare occasions, the engine does not try to pull fuel flow too high. When this occurs, the fuel flow will rise to approximately the correct value, the PID (which is tuned for a large overshoot) pulls the flow far below the correct value, and then corrects it to the correct value. This takes time and, when it occurs, adds another ~2 seconds to the time from idle to takeoff power.

 The following graphs show fuel flow when advancing the throttle from idle to approximately 50% N1. (X axis is time in .2 second increments)

With the current PID values:
old

With the new PID values:
new

The results of this project are not a good as I had hoped when starting it. It seems that the Engine Configuration needs more work. Unfortunately, I do not possess the knowledge to do it. Even if I acquired the knowledge, I do not have the engine references that would be required to do it well. I expect that changing the Engine Configuration would also require changes to the FADEC; I do not have the required references for that either. The real solution to the engine problems will have to be done by someone else.

PR #1508, opened 2022-03-28
fix: payload names

rthom91:


dga711:
Thank you!
Will collect this into a PR for a future release.

PR #1512, opened 2022-03-31
CJ4: Implemented IDENT button in NavRadio

grumpy-anonsaba:
Implemented IDENT functionality in the Nav Radio.

Button is initially white until pressed then turns blue for 10 seconds.
This timing is for vPilot.


mattnischan:
I would recommend reading the ident simvar to check for it being active as opposed to doing a timer in the code; the sim already handles turning ident off after 8 seconds, and having conferred with the vPilot developer on this when they implemented it, they just watch the simvar also.

grumpy-anonsaba:
> I would recommend reading the ident simvar to check for it being active as opposed to doing a timer in the code; the sim already handles turning ident off after 8 seconds, and having conferred with the vPilot developer on this when they implemented it, they just watch the simvar also.

Alright I implemented those checks instead.
Ironically TRANSPONDER IDENT doesn't initially get set to 1, but takes a few in game seconds.
I engineered a cheesy solution to bypass this while we wait on TRANSPONDER IDENT to become active.

It's worth noting the transponder mode must be 0 (on) for IDENT to function properly.
This transponder mode is set to 1 by default (STBY) - So I display an error in the FMC if a user tries to IDENT without the transponder actually turned on

dga711:
Thank you!

